PR #25, opened 2017-05-14
Fix get_keys warning

donadigo:
Fixes `hash_table_get_keys` warnings. It is strongly not recommended to initialize class fields in a thread and using them without knowing if they did really initialize at the usage time. This class would do that almost every time since it calls `load_all_sources` not in the constructor but after creating the class object which is doing initializing in a thread. This branch fixes it.

### DIFF
--- a/src/Widgets/calendar/CalendarModel.vala
+++ b/src/Widgets/calendar/CalendarModel.vala
@@ -130,11 +130,41 @@ namespace DateTime.Widgets {
             lock (calendar_model) {
                 if (calendar_model == null) {
                     calendar_model = new CalendarModel ();
-                    calendar_model.load_all_sources ();
                 }
             }
 
             return calendar_model;
+        }
+
+        construct {
+            this.month_start = Util.get_start_of_month ();
+            compute_ranges ();
+
+            source_client = new HashTable<string, E.CalClient> (str_hash, str_equal);
+            source_events = new HashTable<E.Source, Gee.TreeMap<string, E.CalComponent> > (Util.source_hash_func, Util.source_equal_func);
+            source_view = new HashTable<string, E.CalClientView> (str_hash, str_equal);
+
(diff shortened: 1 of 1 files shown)


parnold-x:
mhh one of the sync method needs some time not sure which one. Did you check for a delay?

donadigo:
@parnold-x yes, I did check what method needs time, it looks like it was `Process.spawn_command_line_sync`. I've moved only initialization of fields, not this one.

parnold-x:
@donadigo nope it was one of the eds functions.  You did move everything eds related from the thread. Do you have data in eds? Or they speed up their initialization.

donadigo:
@parnold-x unfortunately I don't think I have any data in EDS and the only thing I see that could take some time to initialize is `registry = new E.SourceRegistry.sync (null);` but since `registry` is a public property and is used elsewhere in the class, making it threaded doesn't make sense to me, there is also a `foreach (var source in registry.list_sources (E.SOURCE_EXTENSION_CALENDAR))` which adds some sources, this could be put into a thread if needed.

parnold-x:
the purpose to make it threaded was that it took 2-3s. That was the delay on startup, so it was needed...

donadigo:
@parnold-x should we revert it then?

parnold-x:
Your call. Wanted to just clearify the original purpose of the threaded init.
To fix the warning is it enough to just do the hashtable inits in the construct?

donadigo:
Yes, it is enough to do inits of HashTable's, but as I said, putting other fields into treaded initialization could create some problems in the future. We should make sure that these get initted before we use these. I can propose a merge for reverting this and initting only the HashTable's if you want.

parnold-x:
Would be better because a 3s lag at start is significant.

parnold-x:
looks good. Can't approve.

PR #26, opened 2017-05-14
Revert #25

donadigo:
Reverts and adjusts to @parnold-x suggestions in the #25. I've moved everything but the HashTables to the Thread initialization because of possible performance issues.


parnold-x:
Looks good to me.

PR #39, opened 2017-07-12
wrap event labels

danrabbit:
because long event labels make the indicator look dumb AF


jeremypw:
You could maybe get the approximate char width from the label's PangoLayout and use that to calculate the maximum chars from the main grid width but it hardly seems worth the complexity.  Unless someone knows an easier way ...

PR #41, opened 2017-07-25
Synchronize the Model with Maya: use libraries instead of another process

tintou:


src/Widgets/calendar/CalendarModel.vala
```diff
@@ -140,81 +140,40 @@ namespace DateTime.Widgets {
...
-                case 7:
-                    week_starts_on = Weekday.SATURDAY;
-                    break;
-                default:
-                    week_starts_on = Weekday.MONDAY;
-                    message ("Locale has a bad first_weekday value");
-                    break;
+            int week_start = Posix.nl_langinfo2 (Posix.NLTime.FIRST_WEEKDAY).data[0];
+            if (week_start >= 1 && week_start <= 7) {
+                week_starts_on = (Weekday)week_start-1;
```
danrabbit: spaces between math right? :p

Issue #43, opened 2017-08-04
current master (ab6db27) fails to build due to CMake error (?)


decathorpe:
During compilation, I get the following cryptic error (full absolute paths to build directory shortened to "..." for easier reading):

```
cd .../src && /usr/bin/valac -C -b .../src -d .../src --pkg=wingpanel-2.0 --pkg=glib-2.0>=2.32 --pkg=granite --pkg=libecalendar-1.2 --pkg=libedataserver-1.2 --pkg=libical --pkg=libsoup-2.4 --target-glib=2.32 --vapidir=.../vapi --thread -g --save-temps .../src/Indicator.vala .../src/Widgets/PanelLabel.vala .../src/Widgets/calendar/Calendar.vala .../src/Widgets/calendar/CalendarModel.vala .../src/Widgets/calendar/CalendarView.vala .../src/Widgets/calendar/ControlHeader.vala .../src/Widgets/calendar/Util.vala .../src/Widgets/calendar/Grid.vala .../src/Widgets/calendar/GridDay.vala .../src/Widgets/calendar/Header.vala .../src/Widgets/calendar/WeekLabels.vala .../src/Services/TimeManager.vala .../src/Services/ClockSettings.vala .../src/Services/SettingsManager.vala .../src/config.vala .../src/.../vapi/posix-extension.vapi
warning: --save-temps has no effect when -C or --ccode is set
error: .../src/.../vapi/posix-extension.vapi not found
make[2]: *** [src/CMakeFiles/datetime.dir/build.make:125: src/datetime_valac.stamp] Error 1
```

It looks like the path to the "posix-extension.vapi" file contains the build directory path ("...") *twice*, which results in vala not being able to find the file.

jeremypw:
Master builds correctly for me using the command (assuming no existing build directory):

```
mkdir ./build && cd ./build && cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE="Release" .. && make
```

Where does the command starting ```cd .../src && /usr/bin/valac -C -b .../src -d .../src --pkg=wingpanel-2.0 --pkg=glib-2.0>=2.32 --pkg=granite .........```

quoted in the initial report come from?

decathorpe:
Here's a full build log. The commands are the same as for you, except for some more CMake flags that are set on fedora by default.

[...]

PR #49, opened 2017-10-16
Fix the build system not finding the custom vapi

tintou:
Another possible solution for  #43


jeremypw:
This builds OK, but I am not sure what the problem is since master builds OK for me as well (see comment to linked issue).

PR #53, opened 2017-12-28
Clean up Time Manager Settings and add Seconds while at it (Fix #40)

fabianthoma:
In this PR I'm cleaning up the TimeManager Class and the Accociated Label a bit, by doing the following:

- Porting to gsettings with binding
- Listening to Settings Changes to change display modes
- Removing AM/PM Translation handling (Already part of the Locale)
- Adding ability to show/hide date while refactoring this
- Adding ability to show/hide seconds while refactoring this

PR #65, opened 2018-04-22
Change date format to use gsettings value rather than hard coded string.

errata9:


src/Widgets/PanelLabel.vala
```diff
@@ -57,7 +70,7 @@ public class DateTime.Widgets.PanelLabel : Gtk.Grid {
 
     private void update_labels () {
         /// TRANSLATORS: Date format in the panel following http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format */
-        date_label.set_label (Services.TimeManager.get_default ().format (_("%a, %b %e")));
+        date_label.set_label (Services.TimeManager.get_default ().format (_(dateFormat)));
```
Djaler: I think dateFormat shouldn't be translatable now

errata9: Why wouldn't it be? It's still an strftime string, unless the user enters something else, which would be in their own language.

Of course we could check it's a valid string, but why do that? Any part of the string that's not strftime will be output as entered. Maybe that's what the user wants?

Djaler: You can't translate variable, only constant.

errata9: I'm not sure what you mean. I just tested it with a different language and region and it worked fine.

Djaler: It'll compile and works good. But dynamic variable can not be translated. So calling _() is not necessary.

errata9:
OK so I realized that you can customize the datetime format by just editing the localization file in /usr/share/locale. I didn't notice this until now because there is no .mo file for US English. It's a bit of a pain cause the user has to run msgunfmt/msgfmt to get the .po file, edit and rebuild it, but it works fine without having to change the code.

So the question is, should we just add the .po file for English, rather than changing the datetime-indicator code and schema? Or is it better to have the date-format entry in the schema that can be easily changed with gsettings?

donadigo:
@errata9 I think we should still aim for providing that in GSchema, copying the .mo file is kind of a hack that I wouldn't expect to do here.

errata9:
Since this PR conflicts with PR #53 , should I close this one and add the changes there, so both date and time formats can be handled consistently? That way everyone can vote up or down on the whole thing.

fabianthoma:
PR #53 was updated by me to use the Granite implementation of date handling. This should make it consistent with other elementary apps. If there are still changes necessary it's best to propose them to our Granite Library to make all applications benefit from this.

danrabbit:
Superseded by #81, #82, #83

PR #67, opened 2018-05-09
Fixed links of GLib.DateTime.format in Valadoc

ryonakano:
I fixed URL links of GLib.DateTime.format in Valadoc.
The old one may be dead links.


tintou:
Thanks for this :)

PR #70, opened 2018-06-03
Avoid unnecessary nullable uint

jeremypw:


ricotz:
"timeout_id = 0" seems to be missing inside Timeout.add()

jeremypw:
As the next statement reassigns timeout_id, I did not think it necessary - but it could be added if there is a chance of a race.

ricotz:
Then there is no need for remember the timeout_id at all.

PR #81, opened 2018-08-02
Show date setting

danrabbit:
Adds a revealer that hides the date when `org.gnome.interface.clock-show-date` is false


jeremypw:
Date-time settings plug needs a way of changing this setting?  Or is there another way of doing that?

danrabbit:
@jeremypw Yes, for now it would only be accessible through dconf

PR #84, opened 2018-08-03
Use our own gsettings instead of GNOME's

### LINKED ISSUE #57: ISO 8601 for date/time format
There used to be a `custom-time-format` available option in `com.canonical.indicator.datetime`, but this options seems to be gone

The only option I have is to switch between 12h and 24h clocks. I'm stuck with a date format that is alien to me (weekday, 3-letter month, day), when I'd much prefer to see dates in ISO 8601 format (yyyy-mm-dd hh:mm). This shouldn't be determined by my region (which uses days before months, so the current format is wrong anyway), but by some setting specific for dates.

danrabbit:
Cons:
* Might break some third-party tweak tools

Pros:
* Removes a dependency on GSD
* Don't interfere with other DE's settings on multi-DE installations (like might happen on Fedora or Arch)
* Avoid future possible crashes if GNOME changes their keys from under us
* No need to carry default overrides

Something I'm kind of thinking about is adding a setting for `clock-format` that's ISO 8601 to fix #57. So using our own enum here would allow expanding it past just 12 and 24hr.


jeremypw:
Do we need worry about 3rd party tweak tools?  I think not.

danrabbit:
@jeremypw I agree :)

src/Widgets/PanelLabel.vala
```diff
@@ -39,7 +39,7 @@ public class DateTime.Widgets.PanelLabel : Gtk.Grid {
         add (date_revealer);
         add (time_label);
 
-        var clock_settings = new GLib.Settings ("org.gnome.desktop.interface");
+        var clock_settings = new GLib.Settings ("io.elementary.desktop.wingpanel.datetime");
         clock_settings.bind ("clock-format", this, "clock-format", SettingsBindFlags.DEFAULT);
```
jeremypw: clock_setting name should be "clock-formats" to match schema.

danrabbit: The schema is `clock-format` it's only the enum that is `-formats`

screenshot from 2018-08-08 12 54 41 2x

jeremypw:
Weird - it works if "show seconds" is true but not if false.

jeremypw:
OK - finally tracked the problem down to a mistranslation for a time format from the Granite.DateTime.vala in en_GB.po (!).  Nothing to do with this PR.

PR #88, opened 2018-08-09
Add option for ISO 8601 format

### LINKED ISSUE #57: ISO 8601 for date/time format

danrabbit:
Fixes #57


errata9:
Sometimes the small fixes make me happier than big changes. Thanks everybody.

PR #89, opened 2018-08-09
Distinguish weekends

### LINKED ISSUE #6: Grey out weekends
Makes the Calendar look better

Launchpad Details: [#LP1541639](https://bugs.launchpad.net/bugs/1541639) Cody Garver - 2016-02-03 23:24:51 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43024910-grey-out-weekends?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes #6 

Could use some input from @elementary/ux. Not sure how to avoid the conflict between insensitive days of other months and the weekends here. Open to suggestions.


cassidyjames:
Using the proposed CSS:

screenshot from 2018-08-09 10 51 27

davidmhewitt:
I'd like to throw a spanner in the works and say that there are a lot of countries that have different days as the "weekend"

lainsce:
What if saturdays and sundays had a small red tint to their text?

cassidyjames:
@lainsce I don't think red is appropriate, as that's typically an error state. I know some paper calendars do that, but it feels really out of context in elementary OS to me.

lainsce:
Hmm.. what if it was a blue tint for weekends, and the current day had a grey circle behind it then? 🤔

cassidyjames:
@lainsce I think that calls attention to the weekends more than de-prioritizes them, which was the original ask from the issue report.

lainsce:
screenshot from 2018-08-18 08 03 42

How about this for the weekends and the current date? Basically did a 25% less opacity on weekends, and added a circle bg for current date)

jeremypw:
I am unsure about using different shades of grey to denote both 'weekend' and 'insensitive' properties.  There is no shade to denote both properties. 

I would support using color or, better, font for one or other of the properties.

cassidyjames:
Any thoughts from @elementary/ux? I'm kind of at a loss here.

danrabbit:
I'm not really sure what the value of trying to distinguish weekends is here. Isn't it already clear from the column?

cassidyjames:
According to #6, "Makes the Calendar look better" :stuck_out_tongue: I believe I've seen it in other calendars, and I could see it being a little nicer when used for work-week office stuff, but I'm also not really strongly-opinionated about it. Was just trying to find a way to tackle #6. :)

danrabbit:
@cassidyjames I'm kind of inclined to just "won't fix". Seems like more trouble than it's worth in the indicator. Maybe in Calendar proper where you're going to be adding or managing events more?

Issue #4, opened 2017-03-14
Current day blue highlight does not update properly in the calendar


elementaryBot:
When a system has been on over the span of two separate days, the indicator's calendar does not highlight the current day correctly. This is assuming that the day highlighted in blue is supposed to be the "current" day. The attached image shows the indicator in such a state, where the day is the 13th, but the 12th is still highlighted in blue.

Clicking the month above the calendar or moving to a different month seems to trigger an update and the highlight finally moves to the correct day.

Launchpad Details: [#LP1623773](https://bugs.launchpad.net/bugs/1623773) David Berg - 2016-09-15 04:09:05 +0000

elementaryBot:
Cody, why is this high?

Launchpad Details: [#LPC](https://bugs.launchpad.net/wingpanel-indicator-datetime/+bug/1623773/comments/2) Zisu Andrei - 2016-10-26 22:38:43 +0000

cassidyjames:
Marking as high priority because it's actively telling people it's the wrong day in the indicator, which is pretty embarrassing. :grimacing:

peteruithoven:
Looks like we need to call something like [update_day](https://github.com/elementary/wingpanel-indicator-datetime/blob/master/src/Widgets/calendar/Grid.vala#L147-L162) on an timeout similar to the one used in the [TimeManager](https://github.com/elementary/wingpanel-indicator-datetime/blob/master/src/Services/TimeManager.vala) (which is [used](https://github.com/elementary/wingpanel-indicator-datetime/blob/master/src/Widgets/PanelLabel.vala#L54) to update the date and time in the wingpanel).

PR #96, opened 2018-10-10
Update day

peteruithoven:
This should fix: #4

I'm a bit unsure about using format to verify that the date/time changed, in regards to performance. I'm open for suggestions. I could remove this check.


davidmhewitt:
I haven't really read the code to see what your implementation does. But shouldn't it be as simple as updating the current day every time the indicator is opened (I.e. hooking up the indicator opened signal).

There shouldn't really be any performance concerns behind doing that.

peteruithoven:
@davidmhewitt Very good point, I guess I got carried away after refactoring code I didn't like. I've opened a new PR: #97

PR #97, opened 2018-10-10
Update today when calendar is opened

peteruithoven:
Fixes: #4
Supersedes: #96


src/Widgets/calendar/Grid.vala
```diff
@@ -161,6 +161,30 @@ namespace DateTime.Widgets {
...
+            for (i = 0; i < dates.size; i++) {
+                var date = dates[i];
+                GridDay? day = data[day_hash (date)];
+                if (day == null) return;
+                if (date.get_day_of_year () == today.get_day_of_year () && date.get_year () == today.get_year ()) {
+                    day.name = "today";
+                    day.get_style_context ().add_class (Granite.STYLE_CLASS_ACCENT);
+                    day.set_receives_default (true);
+                    day.show_all ();
+                } else if (day.name == "today") {
```
davidmhewitt: It should be possible to move this part of the code that resets the style of dates that aren't today to the `update_day` method above, and then utilise that method in your loop here to save duplicating code between the two methods.

peteruithoven: @davidmhewitt I've now tried to reuse that code in a `update_today_style` function.

PR #98, opened 2018-10-10
Code cleanup

peteruithoven:
Tried making the TimeManager code more readable

### DIFF
--- a/src/Services/TimeManager.vala
+++ b/src/Services/TimeManager.vala
@@ -25,28 +25,23 @@ interface Manager : Object {
 public class DateTime.Services.TimeManager : Gtk.Calendar {
     private static TimeManager? instance = null;
 
-    public signal void minute_changed ();
+    public signal void date_time_changed ();
 
     private GLib.DateTime? current_time = null;
     private uint timeout_id = 0;
     private Manager? manager = null;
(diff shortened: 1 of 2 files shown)


src/Services/TimeManager.vala
```diff
@@ -25,28 +25,23 @@ interface Manager : Object {
...
 
     public TimeManager () {
-        update_current_time ();
-
-        if (current_time == null) {
-            return;
-        }
-
-        add_timeout ();
+        tick();
```
danrabbit: I don't think this is a good method name. `update_current_time` was better imo

peteruithoven: But... I didn't change that name. I've renamed `add_timeout` to `tick` and this function does what should happen every "tick": update time, call the signal, set timeout to be called again. 
One of my issues was that there where already 2 places that had to call: `update_current_time`, `minute_changed` and `add_timeout`.

danrabbit: Er yeah, I guess my point is that `tick` doesn't tell me anything about what this method does. I couldn't even tell what it was replacing :p

peteruithoven:
Alright, giving up on this one, waste of our time.

Issue #95, opened 2018-09-22
Date & Time adds extra space above the calendar


phase1geo:
If I display the Date & Time panel calendar, click on a day that has no events and then switch to a day which has a single event, extra spacing is added above the month selection area as seen in the following image:

![screenshot from 2018-09-22 01 55 44](https://user-images.githubusercontent.com/22555565/45914415-af9c3300-be0a-11e8-89bd-51f5f233e43e.png)

If I switch back to a day which has no events, the extra padding goes away.

J0Ar:
![gifout](https://user-images.githubusercontent.com/23266409/46351006-8a9f8f80-c656-11e8-881e-ae6af393781c.gif)

peteruithoven:
Whoa that was such a weird issue. Apparently it was caused by `add()`ing things to a `Gtk.Grid`.

PR #99, opened 2018-10-10
Fix mystery space

peteruithoven:
Fixes: #95

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -103,15 +103,14 @@ public class DateTime.Indicator : Wingpanel.Indicator {
                 menuitem_label.wrap_mode = Pango.WrapMode.WORD_CHAR;
                 menuitem_label.xalign = 0;
 
-                var menuitem_grid = new Gtk.Grid ();
-                menuitem_grid.column_spacing = 6;
-                menuitem_grid.margin_end = 6;
-                menuitem_grid.margin_start = 6;
-                menuitem_grid.add (menuitem_icon);
-                menuitem_grid.add (menuitem_label);
+                var menuitem_box = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 6);
+                menuitem_box.margin_end = 6;
+                menuitem_box.margin_start = 6;
+                menuitem_box.add (menuitem_icon);
+                menuitem_box.add (menuitem_label);
 
                 var menuitem = new Gtk.Button ();
-                menuitem.add (menuitem_grid);
+                menuitem.add (menuitem_box);
 
                 var style_context = menuitem.get_style_context ();
                 style_context.add_class (Gtk.STYLE_CLASS_MENUITEM);


jeremypw:
Is this a bug in GtkGrid?  The current policy seems to be to prefer GtkGrid as a widget layout container.

src/Indicator.vala
```diff
@@ -49,19 +49,18 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
 
             main_grid = new Gtk.Grid ();
             main_grid.halign = Gtk.Align.CENTER;
             main_grid.valign = Gtk.Align.START;
             main_grid.attach (calendar, 0, 0);
-            main_grid.attach (event_box, 0, 1);
             main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
             main_grid.attach (settings_button, 0, 3);
 
+            create_event_grid ();
```
jeremypw: You do not need this line - the event_box just gets destroyed again.

src/Indicator.vala
```diff
@@ -82,15 +81,25 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
     private void update_events_model (E.Source source, Gee.Collection<E.CalComponent> events) {
         Idle.add (update_events);
     }
 
     private bool update_events () {
-        foreach (var w in event_box.get_children ()) {
-            w.destroy ();
-        }
-        foreach (var e in Widgets.CalendarModel.get_default ().get_events (calendar.selected_date)) {
+        event_grid.destroy();
```
jeremypw: You need to check event_grid exists (non-null) here.  If you move between days with no events twice in a row it will be null on the second function call.

src/Indicator.vala
```diff
@@ -82,15 +81,25 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
     }
 
+    private void create_event_grid () {
+        event_grid = new Gtk.Grid ();
+        event_grid.orientation = Gtk.Orientation.VERTICAL;
+        main_grid.attach (event_grid, 0, 1);
+    }
+
     private void update_events_model (E.Source source, Gee.Collection<E.CalComponent> events) {
         Idle.add (update_events);
```
jeremypw: To prevent this being called twice in different signal handlers, you can create a private uint variable to hold the source id and check whether it is non-zero before creating the Idle.  If it is, remove the pending source and create a new one so that only the last call actually does the update.  Set the source id back to zero at the end of the idle closure.

To DRY the code the checking and creation of the Idle should be in a separate function.  I am not entirely sure the Idle is necessary.  I removed it and connected all the signal handlers that use it with `connect_after` and did not notice a regression.  However, it may be safer to keep it.

I thought that update_events being called concurrently in different threads might have been causing the problem but unfortunately preventing that did not fix it.  I think it is just a bug/feature of Grid.

peteruithoven:
Thanks for the review, I've implemented your suggestions.

src/Indicator.vala
```diff
@@ -81,23 +80,27 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
+    private void update_events_model (E.Source source, Gee.Collection<E.CalComponent> events) {
+        idle_update_events ();
     }
 
-    private void update_events_model (E.Source source, Gee.Collection<E.CalComponent> events) {
-        Idle.add (update_events);
+    private void idle_update_events () {
+        if (update_events_idle_source > 0) {
+            Source.remove (update_events_idle_source);
+            update_events_idle_source = 0;
```
jeremypw: This line is unnecessary as the variable is reassigned in the next line.

src/Indicator.vala
```diff
@@ -81,23 +80,27 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
+        }
+        update_events_idle_source = Idle.add (update_events);
```
jeremypw: Code style:  Blank line after if clause.

src/Indicator.vala
```diff
@@ -81,23 +80,27 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
     }
 
     private bool update_events () {
-        event_grid.destroy();
+        if (event_grid != null) event_grid.destroy();
```
jeremypw: Code style: No single line if clauses without braces. Always use braces and multiple lines.

peteruithoven: We should clarify this at: https://elementary.io/docs/code/reference#reference

peteruithoven:
I've fixed the code style issues.

src/Indicator.vala
```diff
@@ -87,13 +87,15 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
-            update_events_idle_source = 0;
         }
+
         update_events_idle_source = Idle.add (update_events);
     }
 
     private bool update_events () {
-        if (event_grid != null) event_grid.destroy();
+        if (event_grid != null) {
+            event_grid.destroy();
```
jeremypw: Sorry - I missed the lack of a space before the bracket here :-(

Sorry to be fussy.

peteruithoven:
NP, thanks for the check. I should really look into getting the linter working again.

Issue #9, opened 2017-03-14
Shows "undef" text


elementaryBot:
I haven't test it yet but it is possible that the panel label and the today_button will show 'undef' as text.

in  http://bazaar.launchpad.net/~wingpanel-devs/wingpanel-indicator-datetime/trunk/view/head:/src/Services/TimeManager.vala#L48

 public string format (string format) {
  if (current_time == null)
   return "undef";

  return current_time.format (format);
 }

in  the following links you can see this is not catched.
This is not a text you want to show the user

http://bazaar.launchpad.net/~wingpanel-devs/wingpanel-indicator-datetime/trunk/view/head:/src/Widgets/PanelLabel.vala#L26
http://bazaar.launchpad.net/~wingpanel-devs/wingpanel-indicator-datetime/trunk/view/head:/src/Indicator.vala#L90

Launchpad Details: [#LP1467253](https://bugs.launchpad.net/bugs/1467253) kay van der Zander - 2015-06-21 12:53:28 +0000

elementaryBot:
What's the expected behaviour here..? Just showing nothing might cause confusion and then the "my datetime indicator disappeared" bugs are back.. :D

Launchpad Details: [#LPC](https://bugs.launchpad.net/wingpanel-indicator-datetime/+bug/1467253/comments/1) Marcus Wichelmann - 2015-07-13 18:33:59 +0000

fabianthoma:
Hmm I kind of agree here that this is expected behaviour. If we can't fetch the time we can't display it and thus must show a value that represents not getting a time. Maybe we could replace it with a nicer string, but that safeguard should be there.

danrabbit:
I'm gonna close this since there hasn't been any movement on the related PR in 7 months and the issue seems highly theoretical

PR #100, opened 2018-10-10
Replaces undef

peteruithoven:
Fixes: #9
Blocked by string freeze


danrabbit:
Looks like this needs a look over by @elementary/ux

cassidyjames:
Does this mean, "Can't get the local time." will show as the time when we can't get the time? I honestly think just not showing the clock may be better…

cassidyjames:
Whoops, wrong button.

peteruithoven:
Yeah, if the following results in null, 
```
var local_time = new GLib.DateTime.now_local ();
```
which is probably an edge case. you get: 
screenshot from 2018-11-03 10 09 30
screenshot from 2018-11-03 10 09 33

One benefit of this is that it's probably clearer for the user. 

But indeed we could also just hide the indicator. That's what we usually do in indicators and plugs when hardware isn't found.

cassidyjames:
Revisiting this: can we fallback to _some_ sort of time instead of taking away the clock completely?

leonardo-lemos:
Is it possible to fallback to the hardware time?

jeremypw:
How likely is it that this DateTime function fails? The issue seems more theoretical than an actual bug.  I suppose it might fail if (a) the system time is unavailable or invalid or (b) the timezone is unavailable or invalid. Neither seem likely.  Maybe we could fall back to trying to show a DateTime constructed from get_monotonic_time () and the local timezone (or if the timezone is invalid use utc)?

danrabbit:
I'm gonna close this since it's had conflicts for over a year

PR #101, opened 2018-11-04
move "clock-format" GSettings key to granite

decathorpe:
See https://github.com/elementary/granite/pull/242


danrabbit:
This is blocked from merging until those settings are present in a new released package that we can depend on

PR #103, opened 2018-11-11
Focus today on opened

peteruithoven:
Selects / focuses in on today when opened. 
Fixes: https://github.com/elementary/wingpanel-indicator-datetime/issues/102

### DIFF
--- a/src/Widgets/calendar/Grid.vala
+++ b/src/Widgets/calendar/Grid.vala
@@ -69,6 +69,24 @@ namespace DateTime.Widgets {
             }
         }
 
+        public void set_focus_to_today () {
+            debug ("set_focus_to_today");
+            if (grid_range == null) return;
+            debug ("  found grid_range");
+            Gee.List<GLib.DateTime> dates = grid_range.to_list ();
+            int i = 0;
+            for (i = 0; i < dates.size; i++) {
+                var date = dates[i];
+                GridDay? day = data[day_hash (date)];
+                if (day != null && day.name == "today") {
+                    debug ("  found today");
+                    debug ("  can_focus: %s", day.can_focus? "true" : "false");
+                    day.grab_focus_force ();
+                    return;
+                }
+            }
+        }
+
         /**
          * Sets the given range to be displayed in the grid. Note that the number of days
          * must remain the same.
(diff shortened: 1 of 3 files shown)


src/Widgets/calendar/Grid.vala
```diff
@@ -69,6 +69,24 @@ namespace DateTime.Widgets {
             }
         }
 
+        public void set_focus_to_today () {
+            debug ("set_focus_to_today");
+            if (grid_range == null) return;
```
jeremypw: Elementary code style:  No single line if clauses.

src/Widgets/calendar/Grid.vala
```diff
@@ -69,6 +69,24 @@ namespace DateTime.Widgets {
...
+            debug ("  found grid_range");
+            Gee.List<GLib.DateTime> dates = grid_range.to_list ();
+            int i = 0;
```
jeremypw: `i` can be declared in the for statement as only used inside the loop. 
```
for (int i = 0; i < dates.size; i++) {
```

peteruithoven:
Thanks for the review. Yeah this code could use some serious refactoring, especially data flow wise. But I'm kind of afraid of date related stuff.

jeremypw:
You can lose some lines by using the loop in GridDay:set_range () and checking the days to see if they are 'today' there instead of creating a new loop. You will need to grab_focus for the appropriate day in an Idle loop. You can also avoid the grab_focus_force function by making the 'grab_focus' member into an accessible property.

jeremypw:
There is a question over what should happen when you navigate away from the current month and then back.  Should 'today' be reselected?  I guess not, in which case your solution is better as it only runs on creation of the Calendar View.  Using the set_range () loop would me that 'today' would be selected every time you navigate to the current month.

jeremypw:
I am going to approve this in view of #104 which assumes that events (and the selected day) should not be remembered when navigating between months.

peteruithoven:
> There is a question over what should happen when you  navigate away from the current month and then back.  Should 'today' be  reselected?  I guess not, in which case your solution is better as it  only runs on creation of the Calendar View.  Using the set_range () loop  would me that 'today' would be selected every time you navigate to the  current month.

Yes, it was indeed of my goal to not remember the day when navigating between months.

PR #104, opened 2018-11-11
Reset events on month change

### LINKED ISSUE #105: Returning to a month shows events for previously focused day but that day is no longer focused
* Open the calendar widget
* Select a day that has events - that day has a blue circle (focused)
* Move to another month
* Move back to the initial month
* The events still show but there is no focused day (no blue circle).

peteruithoven:
Fixes #105 

By resetting selected_date.

Events are otherwise not cleared when navigating between months.

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -96,7 +96,7 @@ public class DateTime.Indicator : Wingpanel.Indicator {
         if (event_grid != null) {
             event_grid.destroy ();
         }
-
+        if (calendar.selected_date == null) return false;
         var events = Widgets.CalendarModel.get_default ().get_events (calendar.selected_date);
         if (events.size == 0) return false;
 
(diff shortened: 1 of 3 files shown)


jeremypw:
I did not realise this was already in the pipeline when I created the issue!

peteruithoven:
How do I keep forgetting that rule :man_facepalming:
Thanks for the review.

src/Indicator.vala
```diff
@@ -96,10 +96,13 @@ public class DateTime.Indicator : Wingpanel.Indicator {
         if (event_grid != null) {
             event_grid.destroy ();
         }
-        if (calendar.selected_date == null) return false;
+        if (calendar.selected_date == null){
```
jeremypw: Space missing before brace.

peteruithoven: :man_facepalming: thanks, fixed.

PR #110, opened 2018-12-11
Allow users to specifiy their own custom locale-independent dateformat

wolfposd:
- added custom dateformat setting to conf

### DIFF
--- a/src/Widgets/PanelLabel.vala
+++ b/src/Widgets/PanelLabel.vala
@@ -61,21 +61,25 @@ public class DateTime.Widgets.PanelLabel : Gtk.Grid {
 
     private void update_labels () {
         string date_formatString;
-        
-        if(date_format == "custom") {
-        	date_formatString = date_format_custom;
-        }
-        else if (date_format == "ISO8601") {
-        	date_formatString = "%F";
-        }
-        else {
-        	date_formatString = Granite.DateTime.get_default_date_format (clock_show_weekday, true, false);
-        } 
 
-        date_label.label = Services.TimeManager.get_default ().format (date_formatString);
+        if (date_format == "custom") {
+            date_formatString = date_format_custom;
+        } else if (date_format == "ISO8601") {
+            date_formatString = "%F";
+        } else {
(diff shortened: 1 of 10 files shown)


tintou:
The introduction of `ISO8601` as a "clock format" was already mistake, what you want is a datetime format which should be a different key.

wolfposd:
should i create a date format, and use it analogue to the clock-format?

jeremypw:
This works but some validation of user input should be carried out.  If the setting is invalid it should revert to a safe value.

jeremypw:
Maybe the setting description should give more info on how to format the string.

src/Widgets/PanelLabel.vala
```diff
@@ -55,14 +60,19 @@ public class DateTime.Widgets.PanelLabel : Gtk.Grid {
...
 
     private void update_labels () {
-        string date_format;
-        if (clock_format == "ISO8601") {
-            date_format = "%F";
-        } else {
-            date_format = Granite.DateTime.get_default_date_format (clock_show_weekday, true, false);
+        string date_formatString;
+        
+        if(date_format == "custom") {
```
jeremypw: Missing space before '('

src/Widgets/PanelLabel.vala
```diff
@@ -55,14 +60,19 @@ public class DateTime.Widgets.PanelLabel : Gtk.Grid {
...
+        	date_formatString = date_format_custom;
+        }
+        else if (date_format == "ISO8601") {
```
jeremypw: Does not follow elementaryos code style: use ' } else if (...) {'

src/Widgets/PanelLabel.vala
```diff
@@ -55,14 +60,19 @@ public class DateTime.Widgets.PanelLabel : Gtk.Grid {
...
+        	date_formatString = "%F";
         }
+        else {
```
jeremypw: Code style - see above

danrabbit:
Gonna close this since there hasn't been any movement in months and this conflicts with master.

If we want to add more clock format options, we now need to first make them available in the pantheon accountsservice extension provided by `default-settings`

PR #117, opened 2019-03-10
Calendar: don't use absolute path or error silently

### LINKED ISSUE #116: Don't hardcode calendar executable path
It would be much better to use [`GLib.Environment.find_program_in_path`](https://valadoc.org/glib-2.0/GLib.Environment.find_program_in_path.html) for this purpose.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/71172926-don-t-hardcode-calendar-executable-path?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #116 

Throws a dialog on error instead of leaving users in the dark:

Screenshot from 2019-03-09 16 38 01@2x

### DIFF
--- a/src/Widgets/calendar/Calendar.vala
+++ b/src/Widgets/calendar/Calendar.vala
@@ -19,8 +19,6 @@
 
 namespace DateTime.Widgets {
     public class Calendar : Gtk.Box {
-        private const string CALENDAR_EXEC = "/usr/bin/io.elementary.calendar";
-
         ControlHeader heading;
         CalendarView cal;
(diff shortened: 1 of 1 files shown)


worldofpeace:
Nice, different than what i suggested, but it's a great improvement.

##### :crossed_fingers: Now I can drop a downstream patch

PR #120, opened 2019-05-31
Redesign the header of the popup

### LINKED ISSUE #92: Current month in the popup doesn't need the "Go to current Month" button
As it stands, even in the current month, this button is shown:

I propose, that since we're already in the current month, the button isn't there, only shown if the user does change months:
screenshot from 2018-08-18 08 03 42 2

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62379619-current-month-in-the-popup-doesn-t-need-the-go-to-current-month-button?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lainsce:
Fixes #92 

Looks like this:
Screenshot from 2019-07-02 17-36-07

Reviews are more than welcome! :+1:

### DIFF
--- a/data/main.css
+++ b/data/main.css
@@ -1,5 +1,5 @@
 .header-label {
-    font-family: "Raleway", sans;
+    font-family: "Open Sans", sans;
     font-weight: 300;
     font-size: 1.66em;
 }
--- a/data/main.css
+++ b/data/main.css
@@ -1,5 +1,4 @@
 .header-label {
-    font-family: "Open Sans", sans;
     font-weight: 300;
     font-size: 1.66em;
 }
--- a/data/main.css
+++ b/data/main.css
@@ -1,0 +1,5 @@
+.header-label {
+    font-family: "Raleway", sans;
+    font-weight: 300;
+    font-size: 1.66em;
+}
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
     'datetime',
     'vala', 'c',
-    version: '2.1.2'
+    version: '2.1.3'
 )
 
 i18n = import('i18n')
--- a/meson.build
+++ b/meson.build
@@ -12,8 +12,16 @@ add_project_arguments(['--vapidir', join_paths(meson.current_source_dir(), 'vapi
 
 wingpanel_dep = dependency('wingpanel-2.0')
 
+gnome = import('gnome')
+gresource = gnome.compile_resources(
+    'gresource',
+    'data/io.elementary.desktop.wingpanel.datetime.gresource.xml',
+    source_dir: 'data'
+)
+
 shared_module(
     meson.project_name(),
+    gresource,
     'src/DateIterator.vala',
     'src/DateRange.vala',
     'src/Indicator.vala',
--- a/meson.build
+++ b/meson.build
@@ -15,7 +15,7 @@ wingpanel_dep = dependency('wingpanel-2.0')
 gnome = import('gnome')
 gresource = gnome.compile_resources(
     'gresource',
-    'data/io.elementary.desktop.wingpanel.datetime.gresource.xml',
+    join_paths('data', 'io.elementary.desktop.wingpanel.datetime.gresource.xml'),
     source_dir: 'data'
 )
 
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -57,7 +57,7 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             main_grid.hexpand = false;
             main_grid.attach (calendar, 0, 0, 1, 1);
             main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
-            main_grid.attach (settings_button, 0, 3, 1, 1);
+            main_grid.attach (settings_button, 0, 3);
 
             calendar.day_double_click.connect (() => {
                 close ();
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -54,7 +54,6 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             settings_button.text = _("Date & Time Settings…");
 
             main_grid = new Gtk.Grid ();
-            main_grid.hexpand = false;
             main_grid.attach (calendar, 0, 0);
             main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
             main_grid.attach (settings_button, 0, 3);
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -47,8 +47,7 @@ public class DateTime.Indicator : Wingpanel.Indicator {
     public override Gtk.Widget? get_widget () {
         if (main_grid == null) {
             calendar = new Widgets.Calendar ();
-            calendar.margin_top = 6;
-            calendar.margin_bottom = 6;
+            calendar.margin = 6;
 
             var settings_button = new Gtk.ModelButton ();
             settings_button.text = _("Date & Time Settings…");
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -47,11 +47,15 @@ public class DateTime.Indicator : Wingpanel.Indicator {
     public override Gtk.Widget? get_widget () {
         if (main_grid == null) {
             calendar = new Widgets.Calendar ();
+            calendar.margin_top = 6;
+            calendar.margin_bottom = 6;
 
             var settings_button = new Gtk.ModelButton ();
             settings_button.text = _("Date & Time Settings…");
 
             main_grid = new Gtk.Grid ();
+            main_grid.halign = Gtk.Align.CENTER;
+            main_grid.valign = Gtk.Align.START;
             main_grid.attach (calendar, 0, 0);
             main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
             main_grid.attach (settings_button, 0, 3);
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -52,7 +52,6 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             settings_button.text = _("Date & Time Settings…");
 
             main_grid = new Gtk.Grid ();
-            main_grid.margin = 6;
             main_grid.attach (calendar, 0, 0);
             main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
             main_grid.attach (settings_button, 0, 3);
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -54,11 +54,10 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             settings_button.text = _("Date & Time Settings…");
 
             main_grid = new Gtk.Grid ();
-            main_grid.halign = Gtk.Align.CENTER;
-            main_grid.valign = Gtk.Align.START;
-            main_grid.attach (calendar, 0, 0);
-            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
-            main_grid.attach (settings_button, 0, 3);
+            main_grid.hexpand = false;
+            main_grid.attach (calendar, 0, 0, 1, 1);
+            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2, 1, 1);
+            main_grid.attach (settings_button, 0, 3, 1, 1);
 
             calendar.day_double_click.connect (() => {
                 close ();
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -56,7 +56,7 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             main_grid = new Gtk.Grid ();
             main_grid.hexpand = false;
             main_grid.attach (calendar, 0, 0, 1, 1);
-            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2, 1, 1);
+            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
             main_grid.attach (settings_button, 0, 3, 1, 1);
 
             calendar.day_double_click.connect (() => {
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -55,7 +55,7 @@ public class DateTime.Indicator : Wingpanel.Indicator {
 
             main_grid = new Gtk.Grid ();
             main_grid.hexpand = false;
-            main_grid.attach (calendar, 0, 0, 1, 1);
+            main_grid.attach (calendar, 0, 0);
             main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
             main_grid.attach (settings_button, 0, 3);
 
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -47,12 +47,12 @@ public class DateTime.Indicator : Wingpanel.Indicator {
     public override Gtk.Widget? get_widget () {
         if (main_grid == null) {
             calendar = new Widgets.Calendar ();
-            calendar.margin = 6;
 
             var settings_button = new Gtk.ModelButton ();
             settings_button.text = _("Date & Time Settings…");
 
             main_grid = new Gtk.Grid ();
+            main_grid.margin = 6;
             main_grid.attach (calendar, 0, 0);
             main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
             main_grid.attach (settings_button, 0, 3);
--- a/src/Widgets/calendar/Calendar.vala
+++ b/src/Widgets/calendar/Calendar.vala
@@ -31,8 +31,8 @@ namespace DateTime.Widgets {
 
         public Calendar () {
             Object (orientation: Gtk.Orientation.VERTICAL, halign: Gtk.Align.CENTER, valign: Gtk.Align.CENTER, can_focus: false);
-            margin = 6;
-            expand = true;
+            this.margin_start = 10;
+            this.margin_end = 10;
             heading = new ControlHeader ();
             cal = new CalendarView ();
             cal.selection_changed.connect ((date) => {
--- a/src/Widgets/calendar/Calendar.vala
+++ b/src/Widgets/calendar/Calendar.vala
@@ -19,8 +19,6 @@
 
 namespace DateTime.Widgets {
     public class Calendar : Gtk.Box {
-        private const string CALENDAR_EXEC = "/usr/bin/io.elementary.calendar";
-
         ControlHeader heading;
         CalendarView cal;
         public signal void selection_changed (GLib.DateTime? new_date);
@@ -63,14 +61,20 @@ namespace DateTime.Widgets {
 
         // TODO: As far as maya supports it use the Dbus Activation feature to run the calendar-app.
         public void show_date_in_maya (GLib.DateTime date) {
-            var iso_date_string = date.format ("%F");
-            var command = CALENDAR_EXEC + @" --show-day $iso_date_string";
+            var command = "io.elementary.calendar --show-day %s".printf (date.format ("%F"));
 
             try {
                 var appinfo = AppInfo.create_from_commandline (command, null, AppInfoCreateFlags.NONE);
                 appinfo.launch_uris (null, null);
             } catch (GLib.Error e) {
-                warning ("Unable to start calendar, error: %s", e.message);
+                var dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                    _("Unable To Launch Calendar"),
+                    _("The program \"io.elementary.calendar\" may not be installed"),
+                    "dialog-error"
+                );
+                dialog.show_error_details (e.message);
+                dialog.run ();
+                dialog.destroy ();
             }
         }
     }
--- a/src/Widgets/calendar/Calendar.vala
+++ b/src/Widgets/calendar/Calendar.vala
@@ -31,8 +31,8 @@ namespace DateTime.Widgets {
 
         public Calendar () {
             Object (orientation: Gtk.Orientation.VERTICAL, halign: Gtk.Align.CENTER, valign: Gtk.Align.CENTER, can_focus: false);
-            this.margin_start = 10;
-            this.margin_end = 10;
+            margin_start = 0;
+            margin_end = 0;
             heading = new ControlHeader ();
             cal = new CalendarView ();
             cal.selection_changed.connect ((date) => {
--- a/src/Widgets/calendar/Calendar.vala
+++ b/src/Widgets/calendar/Calendar.vala
@@ -31,10 +31,8 @@ namespace DateTime.Widgets {
 
         public Calendar () {
             Object (orientation: Gtk.Orientation.VERTICAL, halign: Gtk.Align.CENTER, valign: Gtk.Align.CENTER, can_focus: false);
-            margin_top = 6;
-            margin_bottom = 6;
-            margin_start = 12;
-            margin_end = 12;
+            margin = 6;
+            expand = true;
             heading = new ControlHeader ();
             cal = new CalendarView ();
             cal.selection_changed.connect ((date) => {
--- a/src/Widgets/calendar/Calendar.vala
+++ b/src/Widgets/calendar/Calendar.vala
@@ -31,8 +31,10 @@ namespace DateTime.Widgets {
 
         public Calendar () {
             Object (orientation: Gtk.Orientation.VERTICAL, halign: Gtk.Align.CENTER, valign: Gtk.Align.CENTER, can_focus: false);
-            margin_start = 0;
-            margin_end = 0;
+            margin_top = 6;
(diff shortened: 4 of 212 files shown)


jeremypw:
Needs input from @elementary/ux, I think. There is an issue of discoverability as it is not obvious what the center button does anyway (there is no tooltip) therefore the change in appearance is unexpected.

lainsce:
I could add a tooltip, but its contents should be decided beforehand by them.

cassidyjames:
@lainsce in Calendar we use "Go to today’s date". This also focuses _today_, not just the current month. However, I think we want to look at reworking the design of this popover anyway, as it's not obvious that clicking a month will focus today, anyway. :stuck_out_tongue:

lainsce:
I've repurposed this PR for this redesign which also fixes the bug as stated in the first comment..

src/Indicator.vala
```diff
@@ -54,11 +54,10 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
 
             main_grid = new Gtk.Grid ();
-            main_grid.halign = Gtk.Align.CENTER;
-            main_grid.valign = Gtk.Align.START;
-            main_grid.attach (calendar, 0, 0);
-            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
-            main_grid.attach (settings_button, 0, 3);
+            main_grid.hexpand = false;
+            main_grid.attach (calendar, 0, 0, 1, 1);
+            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2, 1, 1);
```
danrabbit: You don't need to add width/height here as they're assumed to be 1

```suggestion
            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
```

src/Indicator.vala
```diff
@@ -54,11 +54,10 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
+            main_grid.attach (settings_button, 0, 3, 1, 1);
```
danrabbit: You don't need to add width/height here as they're assumed to be 1
```suggestion
            main_grid.attach (settings_button, 0, 3);
```

src/Indicator.vala
```diff
@@ -54,11 +54,10 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             settings_button.text = _("Date & Time Settings…");
 
             main_grid = new Gtk.Grid ();
-            main_grid.halign = Gtk.Align.CENTER;
-            main_grid.valign = Gtk.Align.START;
-            main_grid.attach (calendar, 0, 0);
-            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
-            main_grid.attach (settings_button, 0, 3);
+            main_grid.hexpand = false;
+            main_grid.attach (calendar, 0, 0, 1, 1);
```
danrabbit: You don't need to add width/height here as they're assumed to be 1
```suggestion
            main_grid.attach (calendar, 0, 0);
```

src/Indicator.vala
```diff
@@ -54,11 +54,10 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             settings_button.text = _("Date & Time Settings…");
 
             main_grid = new Gtk.Grid ();
-            main_grid.halign = Gtk.Align.CENTER;
-            main_grid.valign = Gtk.Align.START;
-            main_grid.attach (calendar, 0, 0);
-            main_grid.attach (new Wingpanel.Widgets.Separator (), 0, 2);
-            main_grid.attach (settings_button, 0, 3);
+            main_grid.hexpand = false;
```
danrabbit: hexpand is false by default, so no need for this line

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -25,30 +25,54 @@ namespace DateTime.Widgets {
...
         public ControlHeader () {
             Object (orientation : Gtk.Orientation.HORIZONTAL);
-            var left_button = new Gtk.Button.from_icon_name ("pan-start-symbolic");
-            var right_button = new Gtk.Button.from_icon_name ("pan-end-symbolic");
-            var center_button = new Gtk.Button.with_label (new GLib.DateTime.now_local ().format (_("%OB %Y")));
+
+            var label = new Gtk.Label (new GLib.DateTime.now_local ().format (_("%OB, %Y")));
+            label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
+            label.halign = Gtk.Align.START;
+            label.set_size_request (185, 32);
```
danrabbit: Where do these numbers come from? Is it not possible to get the desired style without magic numbers?

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -25,28 +25,51 @@ namespace DateTime.Widgets {
...
+            label.halign = Gtk.Align.START;
+            label.set_size_request (185, 24);
+            label.xalign = 0;
+
+            var left_button = new Gtk.Button.from_icon_name ("go-previous-symbolic");
+            var center_button = new Gtk.Button.from_icon_name ("office-calendar-symbolic");
+            center_button.set_tooltip_text (_("Go to today's date"));
+            var right_button = new Gtk.Button.from_icon_name ("go-next-symbolic");
+
+            var box_buttons = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 0);
```
danrabbit: Please don't use Box. It will be removed in future versions of Gtk. Grid is horizontal by default and would work fine here with `grid.add ()`

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -25,28 +25,51 @@ namespace DateTime.Widgets {
...
         public ControlHeader () {
             Object (orientation : Gtk.Orientation.HORIZONTAL);
-            var left_button = new Gtk.Button.from_icon_name ("pan-start-symbolic");
-            var right_button = new Gtk.Button.from_icon_name ("pan-end-symbolic");
-            var center_button = new Gtk.Button.with_label (new GLib.DateTime.now_local ().format (_("%OB %Y")));
+
+            var label = new Gtk.Label (new GLib.DateTime.now_local ().format (_("%OB, %Y")));
+            label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
+            label.halign = Gtk.Align.START;
+            label.set_size_request (185, 24);
```
danrabbit: Where do these numbers come from? Can the desired style be achieved without using magic numbers?

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -25,28 +25,51 @@ namespace DateTime.Widgets {
...
+            label.xalign = 0;
+
+            var left_button = new Gtk.Button.from_icon_name ("go-previous-symbolic");
```
danrabbit: I think in this context, `pan-start-symbolic` and `pan-end-symbolic` probably still look nicer

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -25,28 +25,51 @@ namespace DateTime.Widgets {
...
+            var center_button = new Gtk.Button.from_icon_name ("office-calendar-symbolic");
+            center_button.set_tooltip_text (_("Go to today's date"));
```
danrabbit: It's preferable to use properties where possible for code style
```suggestion
            center_button.tooltip_text = _("Go to today's date");
```

peteruithoven:
I don't really like the left margin of the month and "mon", but with a day selected it kind of makes sense:
Screenshot from 2019-07-02 22-57-28@2x

This margin is especially weird when enabling showing week numbers. 
Screenshot from 2019-07-02 22-55-10@2x
But we probably don't support this officially. It's a setting I've probably customized with dconf-editor, at path: `/io/elementary/desktop/wingpanel/datetime/`. 

Maybe we should make the right margin of the top right buttons consistent with the left margin of the month? 

Why did you change the shape of the blue selected date indicator? (It used to be a circle)

danrabbit:
The weeks numbers setting is exposed in Switchboard, so we definitely want to take it into account

src/Indicator.vala
cassidyjames:
```suggestion
```

src/Widgets/calendar/ControlHeader.vala
lainsce: It's to accomodate big named months without destroying the layout. Don't know of any other way to do this.

lainsce:
The change of the current day indicator is a result of changing the layout, and I can't circle-ify it. Gtk simply does not want to cooperate with me.

peteruithoven:
I guess the choice here is between:
1. We kind of create a large visual left and right margin to allow for the selected day shape. 
You give the calendar a larger left margin when the show-weeks is enabled? That way it would have the same visual left margin as when show-weeks isn't enabled. Combined with using the same margin on the right of the top right buttons and the spacings become quite consistent overall? 
2. We try to keep the left and right margins consistent with the other indicators and accept that that doesn't really align with the calendar part. When show-weeks is enabled the left margins will align though. 

I always noticed that to visually align text you sometimes need to tweak it's position / margins a few pixels, for option 2 I'd move it slightly up for example:
Screenshot from 2019-07-03 01-53-07 2

lainsce:
Try the last commit. It'll make the width wildly jump in size, which is a thing I would like to avoid.

src/Widgets/calendar/ControlHeader.vala
lainsce: Last commit removed magic numbers.

lainsce:
State of Things on last commit:
Screenshot from 2019-07-03 02 21 59

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -25,28 +25,50 @@ namespace DateTime.Widgets {
         public signal void center_clicked ();
         public ControlHeader () {
             Object (orientation : Gtk.Orientation.HORIZONTAL);
+
+            var label = new Gtk.Label (new GLib.DateTime.now_local ().format (_("%OB, %Y")));
+            label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
+            label.halign = Gtk.Align.START;
+            label.set_size_request (185, -1);
```
danrabbit: I think in this case, `width_chars` probably gives you more what you want and is agnostic of the text size in case the user makes it larger or smaller

```suggestion
            label.width_chars = 14;
```

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -25,28 +25,50 @@ namespace DateTime.Widgets {
...
+            box_buttons.add (right_button);
+
             CalendarModel.get_default ().parameters_changed.connect (() => {
                 var date = CalendarModel.get_default ().month_start;
-                center_button.set_label (date.format (_("%OB %Y")));
+                label.set_label (date.format (_("%OB, %Y")));
             });
+
+            var grid = new Gtk.Grid ();
+            grid.orientation = Gtk.Orientation.VERTICAL;
```
danrabbit: Since you're using attach, this is unnecessary

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -25,28 +25,50 @@ namespace DateTime.Widgets {
...
+            grid.column_spacing = 24;
+            grid.margin = 6;
+            grid.attach (label, 0, 0, 1, 1);
```
danrabbit: If you're just packing these next to each other horizontally, why not just use `grid.add ()`?

src/Widgets/calendar/Header.vala
```diff
@@ -35,18 +35,17 @@ public class Header : Gtk.EventBox {
...
         header_grid.margin_bottom = 4;
+        header_grid.hexpand = true;
 
         labels = new Gtk.Label[7];
         for (int c = 0; c < 7; c++) {
             labels[c] = new Gtk.Label ("");
-            labels[c].hexpand = true;
             var label_grid = new Gtk.Grid ();
             label_grid.add (labels[c]);
+            labels[c].hexpand = true;
```
danrabbit: Can you move this back? It doesn't make sense to move it here

src/Widgets/calendar/ControlHeader.vala
lainsce: :+1:

src/Widgets/calendar/ControlHeader.vala
lainsce: I can't change individual items's margins if I use `add ()`. It's dumb.

lainsce:
Lit up the fireworks, because I finally fixed the header's margin woes!
See here:

|With `show-weeks`|Without `show-weeks`|
|:---:|:---:|
|Screenshot from 2019-07-03 22 20 08|Screenshot from 2019-07-03 22 19 49|

lainsce:
Had a small margin calculation error, fixed in last commit.

lainsce:
I think after almost torching the indicator up, I finally got to tame it lmao :joy:
See here:

|With `show-weeks`|Without `show-weeks`|
|:---:|:---:|
|Screenshot from 2019-07-03 23 57 26|Screenshot from 2019-07-03 23 57 42|

lainsce:
State of Things on last commit:
final_result

lainsce:
State of Things after last commit (image provided by @peteruithoven):
d7

jlnr:
I really like the new look, but two things I've noticed:

1. Navigating back and forward with the buttons has a much longer delay now, and this is a beefy machine. Is there some kind of O(n^2) layout going on? (Sorry! I know you've already had a lot of fun with GTK in this branch :sweat_smile:)

2. The buttons are now focusable, they weren't before. I don't think this is a real regression though, because the old focus behavior was a little inconsistent as well (up-arrow would focus the "Date & Time Settings" button).

jeremypw:
You may want to backport the rounded button fix from #126?

lainsce:
Consider it backported.

lainsce:
With the last commit, I consider this PR done for merging.

jlnr:
One minor UX regression from master is that clicking inbetween days immediately closes the popover. I think it should either select the closest day (no gaps between clickable areas), or the popover should ignore the click and stay open.

FYI, the bubbles are not circular with a larger-than-default font size, but that's the case on master as well. I don't think this is worth a bug ticket, but I'll open one for the slowdown issue mentioned above.

lainsce:
State of things after last commit:
Screenshot from 2019-07-06 22 03 02

As you can see, every layout woe has been fixed.

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -19,35 +19,85 @@
...
         public signal void left_clicked ();
         public signal void right_clicked ();
         public signal void center_clicked ();
+
         public ControlHeader () {
             Object (orientation : Gtk.Orientation.HORIZONTAL);
+        }
+
+        construct {
+            string style = """
```
danrabbit: Can you put custom styles in a .css file and load with gresource please?

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -19,35 +19,85 @@
...
+                    font-family: "Raleway", sans;
+                    font-weight: 300;
+                    font-size: 1.66em;
+                }
+            """;
+
+            var provider = new Gtk.CssProvider ();
+            try {
+                provider.load_from_data (style, style.length);
+                Gtk.StyleContext.add_provider_for_screen (Gdk.Screen.get_default (), provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
```
danrabbit: Styles should be applied directly to the widgets, not to the whole screen. Especially because in the panel we don't want styles meant for this indicator to leak into the other indicators

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -19,35 +19,85 @@
...
+            label.xalign = 0;
+            label.width_chars = 13;
+
             var left_button = new Gtk.Button.from_icon_name ("pan-start-symbolic");
+            var center_button = new Gtk.Button.from_icon_name ("office-calendar-symbolic");
+            center_button.tooltip_text = _("Go to today's date");
             var right_button = new Gtk.Button.from_icon_name ("pan-end-symbolic");
-            var center_button = new Gtk.Button.with_label (new GLib.DateTime.now_local ().format (_("%OB %Y")));
+
+            var box_label = new Gtk.Grid ();
```
danrabbit: Why are you adding a new grid that only contains one object?

lainsce: It was part of a dirty hack involving the dummy spacer, but your PR at https://github.com/lainsce/wingpanel-indicator-datetime/pull/4 is a smarter approach.

src/Widgets/calendar/ControlHeader.vala
lainsce: Can do.

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -29,28 +29,20 @@ namespace DateTime.Widgets {
...
-                provider.load_from_data (style, style.length);
-                Gtk.StyleContext.add_provider_for_screen (Gdk.Screen.get_default (), provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+                provider.load_from_resource ("/io/elementary/desktop/wingpanel/datetime/main.css");
+                label.get_style_context ().add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
             } catch (Error e) {
                 critical (e.message);
             }
 
-            var label = new Gtk.Label (new GLib.DateTime.now_local ().format (_("%OB, %Y")));
-            label.hexpand = true;
```
danrabbit: Looks like you lost a property here

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -19,37 +19,64 @@
...
+
+        construct {
+            var provider = new Gtk.CssProvider ();
+
+            var label = new Gtk.Label (new GLib.DateTime.now_local ().format (_("%OB, %Y")));
+            label.get_style_context ().add_class ("header-label");
+            label.xalign = 0;
+            label.width_chars = 13;
+
+            try {
```
danrabbit: You don't need this try/catch anymore since you're loading from resource

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -19,37 +19,64 @@
...
+                provider.load_from_resource ("/io/elementary/desktop/wingpanel/datetime/main.css");
+                label.get_style_context ().add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
```
danrabbit: Instead of getting this style context twice, cache that:

```
            var label_style_context = label.get_style_context ();
            label_style_context.add_class ("header-label");
            label_style_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
```

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -19,37 +19,64 @@
...
         public signal void left_clicked ();
         public signal void right_clicked ();
         public signal void center_clicked ();
+
         public ControlHeader () {
             Object (orientation : Gtk.Orientation.HORIZONTAL);
+        }
+
+        construct {
+            var provider = new Gtk.CssProvider ();
```
danrabbit: Can you group this with `provider.load_from_resource ()`?

src/Widgets/calendar/ControlHeader.vala
lainsce: Better safe than sorry. What if the resource is lost?

danrabbit: It's compiled into the binary, so I guess if part of your binary is missing you have bigger problems :p Anyways, `load_from_resource` doesn't throw an error so this code does nothing: https://valadoc.org/gtk+-3.0/Gtk.CssProvider.load_from_resource.html

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -19,37 +19,61 @@
 
 
 namespace DateTime.Widgets {
-    public class ControlHeader : Gtk.Box {
+    public class ControlHeader : Gtk.Grid {
         public signal void left_clicked ();
         public signal void right_clicked ();
         public signal void center_clicked ();
+
         public ControlHeader () {
```
danrabbit: Since we changed this to a `Gtk.Grid` subclass, these constructor properties are redundant: `Gtk.Grid` is already horizontal by default. So this whole constructor can be removed

data/main.css
```diff
@@ -0,0 +1,5 @@
+.header-label {
+    font-family: "Raleway", sans;
```
danrabbit: Brief chat with @cassidyjames and we agreed that at such a small size, Open Sans looks nicer here. Raleway is only really good as a very large title font and a lot of people really dislike the old fashioned figures

```suggestion
    font-family: "Raleway", sans;
```

src/Widgets/calendar/WeekLabels.vala
```diff
@@ -29,9 +29,10 @@ public class DateTime.Widgets.WeekLabels : Gtk.Revealer {
         vexpand = true;
 
         day_grid = new Gtk.Grid ();
+        day_grid.row_homogeneous = true;
         set_nr_of_weeks (5);
         day_grid.insert_row (1);
-        day_grid.attach (new Gtk.Separator (Gtk.Orientation.VERTICAL), 1, 0, 1, 6);
+        day_grid.attach (new Gtk.Separator (Gtk.Orientation.VERTICAL), 1, 0, 1, nr_of_weeks);
```
danrabbit: Looks like this caused a regression. Best to leave it as is:

Screenshot from 2019-07-08 13 45 39@2x

```suggestion
        day_grid.attach (new Gtk.Separator (Gtk.Orientation.VERTICAL), 1, 0, 1, 6);
```

data/main.css
lainsce: Will do.

data/main.css
```diff
@@ -0,0 +1,5 @@
+.header-label {
+    font-family: "Open Sans", sans;
```
danrabbit: The system default is already Open Sans, so we don't need this line at all
```suggestion
```

meson.build
```diff
@@ -12,8 +12,16 @@ add_project_arguments(['--vapidir', join_paths(meson.current_source_dir(), 'vapi
 
 wingpanel_dep = dependency('wingpanel-2.0')
 
+gnome = import('gnome')
+gresource = gnome.compile_resources(
+    'gresource',
+    'data/io.elementary.desktop.wingpanel.datetime.gresource.xml',
```
danrabbit: join paths:

```suggestion
    join_paths('data', 'io.elementary.desktop.wingpanel.datetime.gresource.xml'),
```

src/Widgets/calendar/Calendar.vala
```diff
@@ -31,8 +31,8 @@ namespace DateTime.Widgets {
 
         public Calendar () {
             Object (orientation: Gtk.Orientation.VERTICAL, halign: Gtk.Align.CENTER, valign: Gtk.Align.CENTER, can_focus: false);
-            this.margin_start = 10;
-            this.margin_end = 10;
+            margin = 6;
+            expand = true;
```
danrabbit: Is this necessary?

src/Widgets/calendar/Header.vala
```diff
@@ -28,25 +28,25 @@ public class Header : Gtk.EventBox {
...
-        header_grid.column_spacing = 0;
-        header_grid.row_spacing = 0;
-        header_grid.margin_bottom = 4;
+        header_grid.set_column_homogeneous (true);
 
         labels = new Gtk.Label[7];
         for (int c = 0; c < 7; c++) {
             labels[c] = new Gtk.Label ("");
-            labels[c].hexpand = true;
+            labels[c].width_chars = 3;
```
danrabbit: This seems unnecessary

src/Widgets/calendar/ControlHeader.vala
```diff
@@ -19,37 +19,57 @@
...
+
+            var box_buttons = new Gtk.Grid ();
+            box_buttons.valign = Gtk.Align.CENTER;
+            box_buttons.get_style_context ().add_class (Gtk.STYLE_CLASS_LINKED);
+            box_buttons.add (left_button);
+            box_buttons.add (center_button);
+            box_buttons.add (right_button);
+
+            column_spacing = 6;
+            margin = 9;
```
danrabbit: This should probably be 6 now to fit in again

```suggestion
            margin = 6;
```

lainsce:
Now it should have zero regressions as well.

PR #123, opened 2019-06-21
Adapt the API to ECal 2.0

tintou:
It's still not fully building with latest libecal-2.0 but it is because of another issue. The branch is also an improvement for the libecal-1.2 version and is using libical for recurrence calculation instead of re-inventing the wheel.

### DIFF
--- a/src/Widgets/calendar/CalendarModel.vala
+++ b/src/Widgets/calendar/CalendarModel.vala
@@ -14,7 +14,6 @@
  *  You should have received a copy of the GNU General Public License
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-
 namespace DateTime.Widgets {
     public class CalendarModel : Object {
         /* The data_range is the range of dates for which this model is storing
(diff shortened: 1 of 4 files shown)


danrabbit:
With this branch it looks like all-day events are off by one. You can see that in git master "Tacos and Tequila Party" is on the 29th (which is correct), but in the new branch, it's on the 30th (incorrect)

Screenshot from 2019-06-21 13 51 42@2x
Screenshot from 2019-06-21 13 52 14@2x

tintou:
@danrabbit I haven't really touched that part, it was already doing this on my end for some events, are you sure that it's a regression with this branch?

danrabbit:
@tintou Yes I compared master and this branch directly

tintou:
@danrabbit can you attach an .ical here so that I have a reproducer

danrabbit:
[All_Day_Test.ics.zip](https://github.com/elementary/wingpanel-indicator-datetime/files/3336788/All_Day_Test.ics.zip)

With master that shows as an all day event on the 29th, but with this branch it shows as being on both the 29th and 30th

danrabbit:
A bigger problem is that it looks like this doesn't show any events that have time spans, only all day events

src/Widgets/calendar/Util.vala
```diff
@@ -80,29 +80,9 @@ namespace Util {
     public TimeZone timezone_from_ical (ICal.Time date) {
         int is_daylight;
         var interval = date.get_timezone ().get_utc_offset (date, out is_daylight);
-        var hours = (interval / 3600);
```
danrabbit: It looks like all this stuff is causing some events to show on the day before

src/Widgets/calendar/CalendarModel.vala
```diff
@@ -121,17 +120,19 @@ namespace DateTime.Widgets {
         /* --- Public Methods ---// */
 
         public Gee.ArrayList<Event> get_events (GLib.DateTime date) {
+            var ical_date = Util.date_time_to_ical (date);
```
danrabbit: It seems like these modifications are the ones that are making it so only all day events show

danrabbit:
Gonna close this since there's a lot of conflicts now

Issue #55, opened 2018-01-20
Deleted events in calendar still show in indicator.


jeremypw:
* Add some events to the current day using the indicator to open calendar.
* Close the calendar and confirm events shown in popup.
* Reopen the calendar and delete the events by editing them and pressing the Delete Event button.
* Close the calendar and look at the popup.  The events are still there.  However, if they are clicked on to open the calendar, they are not in the calendar.

4jNsY6fCVqZv:
Can confirm that this is still an issue with the latest version in Hera!

Instead of restarting Wingpanel, the deleted events disappear from the events column even if I jump one month further/back. Jumping back and forth a day or "Go to today's date" is not enough. Closing and reopening the indicator still shows the events.

One note: The dots under the calendar day disappear immediately. Only the entries in the events column remain.

I would expect the events column to be cleared automatically when the corresponding events are deleted from the calendar.

What could a solution look like?

PR #126, opened 2019-07-04
Add Events Pane header

lainsce:
Adds the events pane header as seen in https://github.com/elementary/mockups/pull/10

Looks like this:
Screenshot from 2019-07-10 00 30 01


lainsce:
State of things on last commit:
Screenshot from 2019-07-05 10 39 21

lainsce:
State of things in last commit:
Screenshot from 2019-07-07 00 41 03

lainsce:
Since I can't put up labels, I'll put one here:

[ Help Wanted ]

jeremypw:
I would suggest that the Events pane make it clearer that the displayed events (or the No events message) are only for the selected day.

lainsce:
Will do, but first, I have Event Dots or Events something to do.

jeremypw:
If you run into difficulties with the event dots you might consider a simpler interim solution so we can merge this PR.  e.g. change the font to bold if events are present.

lainsce:
That's a good idea for this. Since what the PR was intended to do for now is done.

jeremypw:
Yeah - we want to avoid "misson creep" ;-)

lainsce:
I can't seem to get the events to display on the days in the calendar if I don't select them first.

lainsce:
Finally this PR can be merged.

Screenshot from 2019-07-08 21 58 30

lainsce:
This PR is now fit to be merged as soon as #55 is fixed first.

lainsce:
Now this PR should only add in the Events Pane (and associated layout fixes) and nothing more.

src/Indicator.vala
```diff
@@ -47,18 +48,53 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
+            header_grid.column_spacing = 6;
+            header_grid.valign = Gtk.Align.CENTER;
+            header_grid.margin = 6;
+            header_grid.attach (header_label, 0, 0);
+            header_grid.attach (cal_button, 1, 0);
+
+            var sep = new Gtk.Separator (Gtk.Orientation.VERTICAL);
+
+            //TRANSLATORS: Don't remove the "\n\t" part.
+            no_events_label = new Gtk.Label (_("No Events Scheduled\n\tIn This Day"));
```
danrabbit: Definitely don't do this. You can set `max_width_chars` and wrap if your concern is about width

src/Indicator.vala
```diff
@@ -47,18 +48,53 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
+            no_events_label.get_style_context ().add_class (Granite.STYLE_CLASS_H3_LABEL);
+            no_events_label.expand = true;
+            no_events_label.sensitive = false;
```
danrabbit: Please use `Gtk.STYLE_CLASS_DIM_LABEL` instead of setting sensitivity. These are tested for contrast differently

src/Indicator.vala
```diff
@@ -105,33 +149,31 @@ public class DateTime.Indicator : Wingpanel.Indicator {
         var events = Widgets.CalendarModel.get_default ().get_events (calendar.selected_date);
         if (events.size == 0) {
             update_events_idle_source = 0;
+            no_events_label.visible = true;
```
danrabbit: Instead of this you can use ListBox's native placeholder: https://valadoc.org/gtk+-3.0/Gtk.ListBox.set_placeholder.html

src/Widgets/calendar/CalendarModel.vala
```diff
@@ -22,49 +22,59 @@
...
                 alarm = true;
             } else if (Util.is_the_all_day (start_time, end_time)) {
                 day_event = true;
-                return;
             }
+
+            cal = (E.SourceCalendar?)source.get_extension (E.SOURCE_EXTENSION_CALENDAR);
+        }
+
+        public Event (GLib.DateTime date, Util.DateRange range, iCal.Component ical, E.Source source) {
```
danrabbit: Can you make rewriting this class in GObject-style its own PR please?

src/Widgets/calendar/CalendarModel.vala
```diff
@@ -264,92 +300,120 @@ namespace DateTime.Widgets {
...
-            if (wso > 7) {
+            // WSO must be between 1 and 7
+            if (wso > 7)
                 wso = wso - 7;
-            }
 
             offset = 0;
 
-            if (wso < dow) {
+            if (wso < dow)
```
danrabbit: There are a lot of code style regression here. Is this a copy/paste from something?

src/Indicator.vala
lainsce: If I do that, the events pane will never show the `no_events_label` label, even if I rewrite this to be a placeholder. Something awry in the backend, I suppose.

src/Indicator.vala
```diff
@@ -68,6 +107,11 @@ public class DateTime.Indicator : Wingpanel.Indicator {
                 idle_update_events ();
             });
 
+            cal_button.clicked.connect (() => {
+                calendar.open_maya ();
```
danrabbit: This seems to be the only call to this method, so why is it in another class instead of in this class?

src/Indicator.vala
```diff
@@ -89,7 +133,10 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             GLib.Source.remove (update_events_idle_source);
         }
 
-        update_events_idle_source = GLib.Idle.add (update_events);
+        update_events_idle_source = GLib.Idle.add (() => {
+            update_events ();
+            return false;
```
danrabbit: It looks like `update_events` is a boolean method, so why do we return `false` here? It seems like the previous syntax already made sense

src/Widgets/calendar/Calendar.vala
```diff
@@ -19,20 +19,24 @@
 
 namespace DateTime.Widgets {
     public class Calendar : Gtk.Box {
-        ControlHeader heading;
-        CalendarView cal;
+        public ControlHeader heading;
```
danrabbit: Are you sure these need to be public and not private?

src/Widgets/calendar/Util.vala
```diff
@@ -20,6 +20,16 @@
  */
 
 namespace Util {
+    public enum Weekday {
```
danrabbit: What's the reason for moving this from CalendarModel to Util?

src/Widgets/calendar/Util.vala
```diff
@@ -150,6 +160,36 @@ namespace Util {
...
+                            background-color: alpha(%s, 0.15);
+                            color: shade(%s, 0.65);
+                            border-radius: 4px;
+                            box-shadow: 0 1px 2px alpha (#000, 0.2),
+                                        0 1px 3px alpha (%s, 0.15);
+                        }
+                        .%s image {
+                            color: shade(%s, 0.65);
+                        }
+                       """.printf(css_class, color, color, color, css_class, color);
```
danrabbit: Instead of inserting `color` 4 times, make a variable (using `@define-color`) and only insert it once.

src/Widgets/calendar/Calendar.vala
lainsce: Only `ControlHeader heading;` should be, fixing it.

src/Widgets/calendar/Util.vala
lainsce: It's used on another method as being in Util, and it was spitting actual errors out. So I moved it.

lainsce:
As it stands, somewhere along the mergers, it stopped the popover from happeing and crashes Wingpanel.

lainsce:
Finally cleaned up this PR to its minimum.

src/Indicator.vala
```diff
@@ -111,6 +145,25 @@ public class DateTime.Indicator : Wingpanel.Indicator {
         return main_grid;
     }
 
+    // TODO: As far as maya supports it use the Dbus Activation feature to run the calendar-app.
+    public void open_maya () {
```
jlnr: Calling it Maya is a little confusing for us newbies who only know it as "Calendar" :)

lainsce: It's a copypaste from the method in Calendar.vala, adapted to open Calendar. Maya is Calendar's old internal name. A new PR from me I'll rename these. :smile:

danrabbit:
I'm not in favor of adding this header. I feel like it's obvious that this pane contains events and since clicking an event and double clicking a day both launch Calendar, I don't see why we'd want to have a launcher here.

lainsce:
If I remove the launcher, and keep the header, would that be ok?
I wanted to keep both sides with a common layout of header atop content.

cassidyjames:
Would it be absolutely bonkers to have the right pane in a revealer, and collapse it when there are no events? It might get a bit unwieldy when rapidly toggling between days, though.

cassidyjames:
I agree the shortcut to Calendar is not necessary. I also agree that those being events is pretty obvious, but I'm also not sure how to solve the layout without the header.

jlnr:
What about having the events list as a no-delay tooltip for the days in the grid? No clutter by default, less clicks, but all the information is there.

cassidyjames:
As discussed in Slack, dropping the header looks fine.

Screenshot from 2019-07-15 13 29 39@2x
Screenshot from 2019-07-15 13 32 32@2x

lainsce:
Due to the comment above, I'm closing the PR.

PR #135, opened 2019-07-11
Add Event Dots

### LINKED ISSUE #91: Events in the popup calendar should appear as dots
As it stands, the current behavior for showing upcoming events in the month:

I propose then, that days with events should have a dot, maybe red-colored, or not, up to 4 dots:
screenshot from 2018-08-18 08 03 42 1

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62379547-events-in-the-popup-calendar-should-appear-as-dots?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lainsce:
Fixes #91.

Adds the events dots as seen in https://github.com/elementary/mockups/pull/10

Looks like this (after #126 is also merged):
Screenshot from 2019-07-10 00 30 01


src/Widgets/calendar/CalendarModel.vala
```diff
@@ -22,49 +22,55 @@
...
-                return;
             }
+
+            this.cal = (E.SourceCalendar?)source.get_extension (E.SOURCE_EXTENSION_CALENDAR);
         }
 
         public string get_label () {
             if (day_event) {
-                return summary;
+                return "%s\n%s".printf (summary, _("All Day"));
```
danrabbit: Please revert all these string changes for this branch. This doesn't have anything to do with the event dots

src/Widgets/calendar/GridDay.vala
```diff
@@ -39,6 +37,8 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
      */
     public signal void on_event_add (GLib.DateTime date);
 
+    public Gtk.Grid main_grid;
```
danrabbit: Does this need to be public? It seems like other classes shouldn't need to access this

src/Widgets/calendar/CalendarModel.vala
lainsce: They're from the Events Pane PR, was a mis-copypaste, fixing!

src/Widgets/calendar/GridDay.vala
lainsce: Will make it private.

lainsce:
Now it ought to be less laggy when showing the dots and as a consequence, not hang the popover.

src/Widgets/calendar/GridDay.vala
```diff
@@ -82,6 +89,38 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
     public void update_date (GLib.DateTime date) {
         this.date = date;
         label.label = date.get_day_of_month ().to_string ();
+
+        Widgets.CalendarModel.get_default ().events_added.connect (update_event_days);
```
danrabbit: Is there a reason you're adding these here instead of in the construct block? Since this is a public method, this means these signals will possibly be reconnected over and over again

src/Widgets/calendar/GridDay.vala
```diff
@@ -82,6 +89,38 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+            var event_dot_grid = new Gtk.Grid ();
+            event_dot_grid.column_homogeneous = true;
+            if (event_dot_grid != null) {
+                event_dot_grid.destroy ();
+            }
+            if (events.size != 0) {
+                foreach (var e in events) {
+                    if (e != null) {
+                        var event_dot = new Gtk.Image.from_icon_name ("pager-checked-symbolic", Gtk.IconSize.MENU);
+                        event_dot.visible = true;
```
danrabbit: Is this necessary?

danrabbit:
This gets pretty messy if you have a lot of events on one day. Perhaps we should cap it to 3 or 4:

Screenshot from 2019-07-11 12 37 26@2x


I think we want to make sure that the day numbers all line up:

Screenshot from 2019-07-11 12 41 02@2x

src/Widgets/calendar/GridDay.vala
lainsce: Yes, without it, the dot doesn't show. I've tried it without those:
* The icon used is pager-checked-symbolic, to actually draw the dot.
* It's set to `visible = true`, because for some reason it's set to false at Gtk drawing startup or something.

src/Widgets/calendar/GridDay.vala
lainsce: This method is used in `Grid.vala` for the construction of each day. The idea here was, to connect to the signals for each day, and see if they have events or not and display the dots if they have events.

lainsce:
Will make sure to line up the days, and formulate a way to curb to 3 events max.

src/Widgets/calendar/GridDay.vala
lainsce: Nevermind the above, I am dumb.
It's now moved to the construction block and works. Sorry.

lainsce:
Would it make sense UI-wise if a day with more than 4 events had a bar without color below it, signifying "There's events here, and it's more than we can show at once.", like in day 11 here:
Screenshot from 2019-07-11 17 32 11

src/Widgets/calendar/GridDay.vala
```diff
@@ -70,20 +69,70 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+                            event_dot.pixel_size = 6;
+                            event_dot.halign = Gtk.Align.CENTER;
+
+                            Util.set_event_calendar_color (e.cal, event_dot);
+
+                            event_dot_grid.add (event_dot);
+                        }
+                    }
+                } else if (events.size > 4) {
+                    var event_bar = new Gtk.Image.from_icon_name ("list-remove-symbolic", Gtk.IconSize.MENU);
```
danrabbit: This is a very bad idea to abuse icons like this. The design of this icon could change. It would be much better to gresource an icon

src/Widgets/calendar/GridDay.vala
```diff
@@ -70,20 +69,70 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+                    event_bar.visible = true;
+                    event_bar.pixel_size = 6;
```
danrabbit: If you're setting all the same properties, wouldn't it be better to create the image and set its properties first, then set a gicon in the conditional?

src/Widgets/calendar/GridDay.vala
```diff
@@ -70,20 +69,70 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+        add (main_grid);
+        halign = Gtk.Align.CENTER;
         show_all ();
 
         // Signals and handlers
         button_press_event.connect (on_button_press);
         key_press_event.connect (on_key_press);
         scroll_event.connect ((event) => {return Util.on_scroll_event (event);});
+
+        Widgets.CalendarModel.get_default ().events_added.connect (update_event_days);
```
danrabbit: You should probably cache this so you're not getting the calendar model three times in a row

src/Widgets/calendar/CalendarModel.vala
```diff
@@ -16,6 +16,16 @@
  */
 
 namespace DateTime.Widgets {
+    public enum Weekday {
```
danrabbit: Some kind of copy/paste mess up adding this here?

lainsce: It wouldn't compile if this wasn't here.

src/Event.vala
```diff
@@ -21,16 +21,18 @@ public class DateTime.Event : GLib.Object {
...
-
     public GLib.DateTime start_time;
     public bool day_event = false;
+    public E.Source source {get; construct;}
+    public E.SourceCalendar? cal {get; construct;}
 
     private bool alarm = false;
 
-    public Event (GLib.DateTime date, Util.DateRange range, iCal.Component component) {
+    public Event (GLib.DateTime date, Util.DateRange range, iCal.Component ical, E.Source source) {
```
danrabbit: Renaming this has broken this constructor
```suggestion
    public Event (GLib.DateTime date, Util.DateRange range, iCal.Component component, E.Source source) {
```

src/Widgets/calendar/GridDay.vala
```diff
@@ -56,20 +57,65 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+                            Util.set_event_calendar_color (e.cal, event_dot);
+                            event_dot_grid.add (event_dot);
+                        }
+                    }
+                } else if (events.size > 4) {
+                    event_dot.gicon = new ThemedIcon ("events-bar-symbolic");
+                    event_dot_grid.add (event_dot);
+                }
+                event_dot_grid.show_all ();
+                main_grid.margin_top = 5;
```
danrabbit: Instead of adding a margin here, I think setting a height request for the event dot grid would be more reliable

src/Widgets/calendar/GridDay.vala
```diff
@@ -56,20 +57,65 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
 
         label.name = "date";
 
-        add (label);
+        main_grid = new Gtk.Grid ();
+        main_grid.halign = Gtk.Align.CENTER;
+        main_grid.orientation = Gtk.Orientation.VERTICAL;
```
danrabbit: Adding an orientation here doesn't end up doing anything since you're packing widgets using `attach ()` instead of `add ()`

src/Widgets/calendar/GridDay.vala
lainsce: Reliable, but results in this:
Screenshot from 2019-07-12 13 27 20

I'll make some adjustments with Gtk.Align, and make it a non-magic number of 6.

src/Widgets/calendar/GridDay.vala
lainsce: Done.

lainsce:
Supplanted by #163 so closing.

PR #136, opened 2019-07-11
Fix deleted events still showing

### LINKED ISSUE #55: Deleted events in calendar still show in indicator.

lainsce:
Fixes #55

Uses a refresh of `month_start` from `CalendarModel.vala` to fix the bug.


src/Indicator.vala
```diff
@@ -152,6 +152,9 @@ public class DateTime.Indicator : Wingpanel.Indicator {
     }
 
     public override void opened () {
+        Widgets.CalendarModel.get_default ().month_start = Util.get_start_of_month ();
```
jlnr: If this has no effect other than refreshing the events, then I think it would be good to have a comment to that effect above these two lines.

lainsce: Will do.

lainsce:
Nevermind this PR, it's a relic now.

PR #140, opened 2019-07-12
Make event labels smarter

lainsce:
Instead of not showing via their labels, make shown:
- All-Day events
- Alarms
- End times of events that span times

### DIFF
--- a/src/Event.vala
+++ b/src/Event.vala
@@ -26,6 +26,11 @@ public class DateTime.Event : GLib.Object {
     public GLib.DateTime end_time;
     public bool day_event = false;
 
+    // For ListBox use.
+    public string event_times;
+    public string all_day_str;
+    public string alarm_str;
+
     private bool alarm = false;
 
     public Event (GLib.DateTime date, Util.DateRange range, iCal.Component component) {
@@ -44,18 +49,15 @@ public class DateTime.Event : GLib.Object {
         } else if (Util.is_the_all_day (start_time, end_time)) {
             day_event = true;
         }
+
+        event_times = "%s - %s".printf(start_time.format (Util.TimeFormat ()), end_time.format (Util.TimeFormat ()));
+        all_day_str = "%s".printf (_("All Day:"));
+        alarm_str = "%s".printf (_("Alarm:"));
     }
 
     public string get_label () {
         var summary = component.get_summary ();
-        if (day_event) {
-            return "%s\n%s".printf (summary, _("All Day"));
-        } else if (alarm) {
-            return "%s %s\n%s".printf (_("Alarm:"), start_time.format (Util.TimeFormat ()), summary);
-        } else if (range.days > 0 && date.compare (range.first_dt) != 0) {
-            return summary;
-        }
-        return "%s\n%s - %s".printf (summary, start_time.format (Util.TimeFormat ()), end_time.format (Util.TimeFormat ()));
+        return summary;
     }
 
     public string get_icon () {
--- a/src/Event.vala
+++ b/src/Event.vala
@@ -46,12 +46,12 @@ public class DateTime.Event : GLib.Object {
         }
     }
 
-    public void get_event_label (out string summary) {
-        summary = component.get_summary ();
+    public string get_event_label () {
+        return component.get_summary ();
     }
 
-    public void get_event_times (out string event_times) {
-        event_times = "%s - %s".printf(start_time.format (Util.TimeFormat ()), end_time.format (Util.TimeFormat ()));
+    public string get_event_times () {
+        return "%s - %s".printf(start_time.format (Util.TimeFormat ()), end_time.format (Util.TimeFormat ()));
     }
 
     public string get_icon () {
--- a/src/Event.vala
+++ b/src/Event.vala
@@ -44,9 +44,6 @@ public class DateTime.Event : GLib.Object {
     }
 
     public string get_icon () {
-        if (alarm) {
-            return "alarm-symbolic";
-        }
         return "office-calendar-symbolic";
     }
 }
--- a/src/Event.vala
+++ b/src/Event.vala
@@ -51,6 +51,9 @@ public class DateTime.Event : GLib.Object {
     }
 
     public string get_event_times () {
+        if (day_event) {
+            return "";
+        }
         return "%s - %s".printf(start_time.format (Util.TimeFormat ()), end_time.format (Util.TimeFormat ()));
     }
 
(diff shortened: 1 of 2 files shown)


src/Event.vala
```diff
@@ -48,13 +49,13 @@ public class DateTime.Event : GLib.Object {
     public string get_label () {
         var summary = component.get_summary ();
         if (day_event) {
-            return summary;
+            return "%s\n%s".printf (summary, _("All Day"));
         } else if (alarm) {
-            return "%s - %s".printf (start_time.format (Util.TimeFormat ()), summary);
+            return "%s %s\n%s".printf (_("Alarm:"), start_time.format (Util.TimeFormat ()), summary);
```
danrabbit: Please don't put newlines in the middle of strings. Use two labels so we can lay them out accurately

lainsce:
> Instead of doing this, I think it would be smarter (and more consistent) to use listbox headers
That's for another PR, I suppose; these labels inside `get_label ()` are the raw event strings.

lainsce:
In the last commit:
Screenshot from 2019-07-12 14 49 31

src/Event.vala
```diff
@@ -36,25 +37,21 @@ public class DateTime.Event : GLib.Object {
...
-        var summary = component.get_summary ();
-        if (day_event) {
-            return summary;
-        } else if (alarm) {
-            return "%s - %s".printf (start_time.format (Util.TimeFormat ()), summary);
-        } else if (range.days > 0 && date.compare (range.first_dt) != 0) {
-            return summary;
-        }
-        return "%s - %s".printf (summary, start_time.format (Util.TimeFormat ()));
+    public void get_event_label (out string summary) {
```
danrabbit: If this only returns one string, why a void method instead of a string method?

lainsce:
How's the last commit?

src/Event.vala
```diff
@@ -36,25 +37,21 @@ public class DateTime.Event : GLib.Object {
...
+        summary = component.get_summary ();
+    }
+
+    public void get_event_times (out string event_times) {
```
danrabbit: Same here

src/Indicator.vala
```diff
@@ -116,7 +116,11 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             var menuitem_icon = new Gtk.Image.from_icon_name (e.get_icon (), Gtk.IconSize.MENU);
             menuitem_icon.valign = Gtk.Align.START;
 
-            var menuitem_label = new Gtk.Label (e.get_label ());
+            string label, times;
+            e.get_event_label (out label);
+            e.get_event_times (out times);
+            var menuitem_label = new Gtk.Label ("");
+            menuitem_label.set_markup ("<b>%s</b>\n%s".printf(label, times));
```
danrabbit: Please use two labels, not a single label with a newline

lainsce:
All fixed.

lainsce:
That last fix was for this, all-day events:
Screenshot from 2019-07-12 15 03 27

src/Event.vala
```diff
@@ -36,25 +37,24 @@ public class DateTime.Event : GLib.Object {
...
         if (day_event) {
-            return summary;
-        } else if (alarm) {
-            return "%s - %s".printf (start_time.format (Util.TimeFormat ()), summary);
-        } else if (range.days > 0 && date.compare (range.first_dt) != 0) {
-            return summary;
+            return "";
         }
-        return "%s - %s".printf (summary, start_time.format (Util.TimeFormat ()));
+        return "%s - %s".printf(start_time.format (Util.TimeFormat ()), end_time.format (Util.TimeFormat ()));
```
danrabbit: Don't forget spaces before ()

```suggestion
        return "%s - %s".printf (start_time.format (Util.TimeFormat ()), end_time.format (Util.TimeFormat ()));
```

src/Indicator.vala
```diff
@@ -116,7 +116,8 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             var menuitem_icon = new Gtk.Image.from_icon_name (e.get_icon (), Gtk.IconSize.MENU);
             menuitem_icon.valign = Gtk.Align.START;
 
-            var menuitem_label = new Gtk.Label (e.get_label ());
+            var menuitem_label = new Gtk.Label ("");
+            menuitem_label.set_markup ("<b>%s</b>".printf(e.get_event_label ()));
```
danrabbit:
```suggestion
            menuitem_label.set_markup ("<b>%s</b>".printf (e.get_event_label ()));
```

src/Indicator.vala
```diff
@@ -125,11 +126,24 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             menuitem_label.wrap_mode = Pango.WrapMode.WORD_CHAR;
             menuitem_label.xalign = 0;
 
-            var menuitem_box = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 6);
+            var menuitem_times = new Gtk.Label ("");
+            menuitem_times.set_markup (e.get_event_times ());
+            menuitem_times.hexpand = true;
+            menuitem_times.lines = 3;
```
danrabbit: We shouldn't need any of this for times, they'll always fit on one line since they'll be a maximum of like 20 characters

src/Indicator.vala
```diff
@@ -125,11 +126,24 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             menuitem_label.wrap_mode = Pango.WrapMode.WORD_CHAR;
             menuitem_label.xalign = 0;
 
-            var menuitem_box = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 6);
+            var menuitem_times = new Gtk.Label ("");
+            menuitem_times.set_markup (e.get_event_times ());
```
danrabbit: In Calendar we use both `<small>` and `DIM_LABEL`. I think it probably makes sense to do that here as well
```suggestion
            menuitem_times.set_markup ("<small>%s</small>".printf (e.get_event_times ()));
```

src/Indicator.vala
```diff
@@ -125,11 +126,18 @@ public class DateTime.Indicator : Wingpanel.Indicator {
             menuitem_label.wrap_mode = Pango.WrapMode.WORD_CHAR;
             menuitem_label.xalign = 0;
 
-            var menuitem_box = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 6);
+            var menuitem_times = new Gtk.Label ("");
```
danrabbit: You do still need halign :p

lainsce: xalign, halign wouldn't do anything.

PR #143, opened 2019-07-12
Coloring events in the events pane by calendar colors

lainsce:
Should catch events calendar colors and apply to them

Screenshot from 2019-07-12 17 01 27


data/EventRow.css
```diff
@@ -1,10 +1,14 @@
...
-    background: alpha(@colorAccent, 0.15);
-    border-radius: 3px;
-    color: shade(@colorAccent, 0.65);
-    padding: 6px;
+    background-color: alpha(@eventcolorAccent, 0.15);
+    color: shade(@eventcolorAccent, 0.65);
+    border-radius: 4px;
+    padding: 10px;
+    box-shadow: 0 1px 2px alpha (#000, 0.2),
+                0 1px 3px alpha (@eventcolorAccent, 0.15);
```
danrabbit: These styles match what's in Calendar, so I'd prefer not to change this

src/Widgets/EventRow.vala
```diff
@@ -1,18 +1,18 @@
 /*
  * Copyright 2019 elementary, Inc. (https://elementary.io)
  *
- * This program is free software; you can redistribute it and/or
+ * grid program is free software; you can redistribute it and/or
```
danrabbit: find/replace errors

data/EventRow.css
lainsce: At least the padding should be added otherwise the text and icons will hug the border of the box

lainsce:
Uses eventcolorAccent so that the today circle doesn't get colored as well :smile:

data/EventRow.css
```diff
@@ -1,10 +1,12 @@
 .event {
-    background: alpha(@colorAccent, 0.15);
+    background: alpha(@eventcolorAccent, 0.15);
```
danrabbit: Since we add the provider to the widget and not the screen, this is not necessary

data/EventRow.css
```diff
@@ -1,10 +1,12 @@
...
     border-radius: 3px;
-    color: shade(@colorAccent, 0.65);
-    padding: 6px;
+    color: shade(@eventcolorAccent, 0.65);
+    padding: 10px;
```
danrabbit: Please revert this change

data/EventRow.css
```diff
@@ -1,10 +1,12 @@
...
 }
-
 .event .title {
     font-weight: 600;
 }
+.event .icon {
```
danrabbit: You don't need to add a new class here. You can select with the element type: `.event image`

src/Widgets/EventRow.vala
```diff
@@ -37,6 +37,7 @@ public class DateTime.EventRow : Gtk.ListBoxRow {
 
         var event_image_context = event_image.get_style_context ();
         event_image_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+        event_image_context.add_class ("icon");
```
danrabbit: Adding a class isn't necessary here. See above

src/Widgets/EventRow.vala
```diff
@@ -70,6 +71,32 @@ public class DateTime.EventRow : Gtk.ListBoxRow {
...
+
+        cal_event.cal.notify["color"].connect (() => {
+            get_style_calendar_color (cal_event.cal, grid);
+            get_style_calendar_color (cal_event.cal, event_image);
+        });
+
         add (grid);
     }
+
+    public void get_style_calendar_color (E.SourceCalendar cal, Gtk.Widget widget) {
```
danrabbit: Since we're going to want this later for the event dots, we should just put this in Util and save ourselves that work in the other branch of moving it

lainsce:
All done.

src/Widgets/EventRow.vala
```diff
@@ -70,6 +70,15 @@ public class DateTime.EventRow : Gtk.ListBoxRow {
         grid_context.add_class ("event");
         grid_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
 
+        /* Color menuitem per calendar source of event */
+        Util.get_style_calendar_color (cal_event.cal, grid);
+        Util.get_style_calendar_color (cal_event.cal, event_image);
+
+        cal_event.cal.notify["color"].connect (() => {
+            Util.get_style_calendar_color (cal_event.cal, grid);
+            Util.get_style_calendar_color (cal_event.cal, event_image);
```
danrabbit: DRY this please

src/Event.vala
```diff
@@ -21,17 +21,20 @@ public class DateTime.Event : GLib.Object {
     public GLib.DateTime date { get; construct; }
     public unowned iCal.Component component { get; construct; }
     public Util.DateRange range { get; construct; }
+    public E.Source source {get; construct;}
```
danrabbit: Code style
```suggestion
    public E.Source source { get; construct; }
```

src/Widgets/EventRow.vala
lainsce: How do I do it, it should run the method for both, otherwise the icon never gets the color.

lainsce:
Again, all done.

src/Widgets/calendar/CalendarModel.vala
```diff
@@ -121,20 +121,27 @@ namespace DateTime.Widgets {
...
-            var events_on_day = new Gee.TreeMap<string,Event> ();
-            foreach (var entry in source_events.get_values ()) {
-                foreach (var comp in entry.values) {
-                    unowned iCal.Component ical = comp.get_icalcomponent ();
-                    foreach (var dt_range in Util.event_date_ranges (ical, data_range)) {
-                        if (dt_range.contains (date)) {
-                            if (!events_on_day.has_key (ical.get_uid ())) {
-                                events_on_day.set (ical.get_uid (), new Event (date, dt_range, ical));
+            var events = new Gee.TreeMap<string,Event> ();
+            registry.list_sources (E.SOURCE_EXTENSION_CALENDAR).foreach ((source) => {
```
danrabbit: So it kind of looks like we don't need to do this. In the code you deleted, there's a hashmap `source_events` that we're iterating through. So it kind of seems like we can get the source from there instead of having to list all sources again

lainsce: It's a different kind of Source. We want the sources with E.SOURCE_EXTENSION_CALENDAR enabled here. Can't get those from source_events, as those are just the events themselves.

jeremypw:
@lainsce Is this PR still needed - the events in master are now colored correctly?

lainsce:
It's needed since the events pane event box colors are always blue, unless there was a PR doing this.

jeremypw:
Ah, OK - it is only the event dots fixed in the indicator - I got confused with the Calendar app.

src/Event.vala
```diff
@@ -21,15 +21,18 @@ public class DateTime.Event : GLib.Object {
     public GLib.DateTime date { get; construct; }
     public unowned ICal.Component component { get; construct; }
     public Util.DateRange range { get; construct; }
+    public E.Source source { get; construct; }
+    public E.SourceCalendar? cal {get; construct;}
```
jeremypw: Missing spaces

src/Widgets/EventRow.vala
```diff
@@ -90,4 +99,10 @@ public class DateTime.EventRow : Gtk.ListBoxRow {
         var time_format = Granite.DateTime.get_default_time_format (time_manager.is_12h);
         time_label.label = "<small>%s – %s</small>".printf (cal_event.start_time.format (time_format), cal_event.end_time.format (time_format));
     }
+    public void set_color () {
+        var provider = new Gtk.CssProvider ();
+        Util.get_style_calendar_color (cal_event.cal, provider);
+        grid_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+        event_image_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
```
jeremypw: Not sure you should be adding new providers for color without removing the old ones. Why not keep one color provider per instance and just update it?

src/Widgets/calendar/Util.vala
```diff
@@ -20,6 +20,18 @@
  */
 
 namespace Util {
+    public void get_style_calendar_color (E.SourceCalendar cal, Gtk.CssProvider provider) {
+        var color = cal.dup_color ();
+        string style = """
+                        @define-color colorAccent %s;
+                       """.printf(color);
```
jeremypw: Missing space.

src/Widgets/calendar/Util.vala
```diff
@@ -20,6 +20,18 @@
  */
 
 namespace Util {
+    public void get_style_calendar_color (E.SourceCalendar cal, Gtk.CssProvider provider) {
+        var color = cal.dup_color ();
```
jeremypw: Don't really need a var here - its only used once?

src/Widgets/EventRow.vala
lainsce: It's how it works for adding color without using the screen method. :man_shrugging:

jeremypw: It works OK if you have an instance member `color_css_provider` rather than creating a new one when the color changes.  However, I am not sure you need to connect to the `notify["color"]` signal, in practice.  I cannot find a way of changing the color of the calendar without the indicator losing focus and closing. If you do not connect to this signal then the provider only gets added once anyway so you do not need an instance member.

lainsce: The color signal is for e.g. when the user changes the Calendar color externally, so I'm afraid I can't remove it. I will however think about how to implement color_css_provider. A snippet of code would help me immensely.

jeremypw: I still cannot think of a way of changing the calendar color externally while maintaining focus on this calendar indicator. When this indicator loses focus it closes and the color will be updated when it is re-opened. Or are you thinking that the calendar may be accessed from another computer while the indicator is open?  Pretty unlikely, but I suppose it may be possible.  I'll give a code snippet below.

jeremypw:
Code to use instance member provider:

```
public class DateTime.EventRow : Gtk.ListBoxRow {
    public DateTime.Event cal_event { get; construct; }

    private static Services.TimeManager time_manager;
    private static Gtk.CssProvider css_provider;
    private Gtk.StyleContext event_image_context;
    private Gtk.StyleContext grid_context;
    private Gtk.CssProvider color_css_provider;
....
    construct {
        color_css_provider = new Gtk.CssProvider ();
....
    public void set_color () {
        Util.get_style_calendar_color (cal_event.cal, color_css_provider);
        /* Gtk ensures that color_css_provider object is removed first */
        grid_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
        event_image_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
    }
```

src/Widgets/EventRow.vala
lainsce: Fixed.

lainsce:
Should be good to go after fixing jeremypw's comments.

jeremypw:
Not sure why the fix is not needed. Is this because "set_color" is only called once - in which case you do not need to connect to the signal - or because you think Gtk will remove the previous version?  I do not think the latter case is true since you are creating a new object each time so Gtk will not detect it as a duplicate(?)

lainsce:
@jeremypw By adding the fix, you can see it didn't build, so I removed it and it builds.

jeremypw:
It didn't build because you did not add the fix accurately.

lainsce:
Oh, oops.

lainsce:
Let this PR be a reference now, since other PRs merged faster than it.

lainsce:
This PR is back on the menu!

src/Widgets/EventRow.vala
```diff
@@ -20,20 +20,25 @@
...
     public unowned ICal.Component component { get; construct; }
+    public unowned E.SourceCalendar cal { get; construct; }
 
     public GLib.DateTime start_time { get; private set; }
     public GLib.DateTime? end_time { get; private set; }
     public bool is_allday { get; private set; default = false; }
 
     private static Services.TimeManager time_manager;
     private static Gtk.CssProvider css_provider;
+    private static Gtk.CssProvider css_color_provider;
```
jeremypw: This should not be declared  `static` - it is assigned in the instance construct clause.

src/Widgets/EventRow.vala
```diff
@@ -105,4 +118,9 @@ public class DateTime.EventRow : Gtk.ListBoxRow {
         var time_format = Granite.DateTime.get_default_time_format (time_manager.is_12h);
         time_label.label = "<small>%s – %s</small>".printf (start_time.format (time_format), end_time.format (time_format));
     }
+    public void set_color () {
```
jeremypw: Lint: Missing line between functions.

lainsce:
Those two can be done in a new PR

jeremypw:
OK - the PR was merged while I was testing.

PR #147, opened 2019-07-15
Get 12h from TimeManager

danrabbit:


danrabbit:
Gonna put this on hold since we're moving around the namespace of the accountsservice extension

PR #159, opened 2019-07-17
Remove Header class

danrabbit:
Builds on #155 

We only use one header class per big grid and it's such a tiny class it seems like it makes things easier to just combine it


jeremypw:
There seems to be a layout problem with this PR, when `show-week-numbers` setting is false:

Screenshot from 2019-07-30 16 35 47

PR #162, opened 2019-07-18
CalenderModel: remove unused signals and methods

danrabbit:
And appropriately scope remaining variables and methods


jeremypw:
Still compiles and runs after master merge ..

PR #163, opened 2019-07-18
Add dots for events in the calendar

### LINKED ISSUE #91: Events in the popup calendar should appear as dots

danrabbit:
Fixes #91

Screenshot from 2019-07-18 10 59 20@2x

### DIFF
--- a/src/Widgets/calendar/GridDay.vala
+++ b/src/Widgets/calendar/GridDay.vala
@@ -31,10 +31,10 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
     public GLib.DateTime date { get; construct set; }
     public int id { get; construct; }
 
+    private static Gtk.CssProvider provider;
     private static Widgets.CalendarModel model;
 
     private Gee.HashMap<string, Gtk.Widget> event_dots;
-    private Gtk.CssProvider provider;
     private Gtk.Grid event_grid;
     private Gtk.Label label;
     private bool valid_grab = false;
@@ -48,12 +48,12 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
 
     static construct {
         model = Widgets.CalendarModel.get_default ();
+
+        provider = new Gtk.CssProvider ();
+        provider.load_from_resource ("/io/elementary/desktop/wingpanel/datetime/GridDay.css");
     }
 
     construct {
-        provider = new Gtk.CssProvider ();
-        provider.load_from_resource ("/io/elementary/desktop/wingpanel/datetime/GridDay.css");
-
         unowned Gtk.StyleContext style_context = get_style_context ();
         style_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
         style_context.add_class ("circular");
--- a/src/Widgets/calendar/GridDay.vala
+++ b/src/Widgets/calendar/GridDay.vala
@@ -31,16 +31,22 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
     public GLib.DateTime date { get; construct set; }
     public int id { get; construct; }
 
-    private Widgets.CalendarModel model;
+    private static Widgets.CalendarModel model;
     private Gtk.Grid event_grid;
     private Gtk.Label label;
     private bool valid_grab = false;
+
+    private Gee.HashMap<string, Gtk.Widget> event_dots;
 
     public GridDay (GLib.DateTime date, int id) {
         Object (
             date: date,
             id: id
         );
+    }
+
+    static construct {
+        model = Widgets.CalendarModel.get_default ();
     }
 
     construct {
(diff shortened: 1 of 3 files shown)


src/Widgets/calendar/GridDay.vala
```diff
@@ -68,6 +86,55 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+    }
+
+    private void add_event_dots (E.Source source, Gee.Collection<ECal.Component> events) {
+        foreach (var component in events) {
+            if (event_dots.size >= 3) {
+                return;
+            }
+
+            unowned ICal.Component ical = component.get_icalcomponent ();
+            foreach (var dt_range in Util.event_date_ranges (ical, model.data_range)) {
```
donadigo: I wonder if this inner foreach can run many times and exceed the initial `event_dots.size >= 3` constraint? Is this something that we should be concerned with?

src/Widgets/calendar/GridDay.vala
```diff
@@ -68,6 +86,55 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+                if (dt_range.contains (date)) {
```
donadigo:
```suggestion
                if (date in dt_range) {
```

You can use the native Vala syntax for Gee classes.

src/Widgets/calendar/GridDay.vala
```diff
@@ -68,6 +86,55 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+                        event_dot.pixel_size = 6;
+
+                        unowned Gtk.StyleContext style_context = event_dot.get_style_context ();
+                        style_context.add_class (Granite.STYLE_CLASS_ACCENT);
+                        style_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+
+                        var source_calendar = (E.SourceCalendar?) source.get_extension (E.SOURCE_EXTENSION_CALENDAR);
+                        Util.set_event_calendar_color (source_calendar, event_dot);
+
+                        event_dots.set (event_uid, event_dot);
```
donadigo:
```suggestion
                        event_dots[event_uid] = event_dot;
```

src/Widgets/calendar/GridDay.vala
```diff
@@ -68,6 +86,55 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+        }
+
+        event_grid.show_all ();
+    }
+
+    private void remove_event_dots (E.Source source, Gee.Collection<ECal.Component> events) {
+        foreach (var component in events) {
+            unowned ICal.Component ical = component.get_icalcomponent ();
+            var event_uid = ical.get_uid ();
+            if (event_dots.has_key (event_uid)) {
```
donadigo: You may as well get the value directly and then check if it's null:
```
var dot = event_dots[event_uid];
if (dot != null) {
    dot.destroy ();
    event_dots.remove (dot);
}
```

src/Widgets/calendar/Util.vala
```diff
@@ -499,6 +499,24 @@ namespace Util {
...
 
+    public void set_event_calendar_color (E.SourceCalendar cal, Gtk.Widget widget) {
+        var color = cal.dup_color ();
+
+        string style = """
+            @define-color colorAccent %s;
+        """.printf (color);
+
+        try {
+            var style_provider = new Gtk.CssProvider ();
```
donadigo: I think that we could optimize this here by storing providers in a HashMap and then creating a new provider only if the color is new and doesn't exist in the provider list. Otherwise, we just get the existing provider.

src/Widgets/calendar/GridDay.vala
danrabbit: Yeah moving that check seems to actually help with under-reporting the number of dots also

src/Widgets/calendar/Util.vala
```diff
@@ -499,6 +499,32 @@ namespace Util {
         }
     }
 
+    private Gee.HashMap<string, Gtk.CssProvider> providers;
```
donadigo:
```suggestion
    private Gee.HashMap<string, Gtk.CssProvider>? providers;
```

Mark this as nullable.

PR #167, opened 2019-07-18
EventRow: no longer depend on Event

danrabbit:
Makes more sense in the context of #166


jeremypw:
What is the rationale for removing date_range from "Event"? I guess since only one day is shown at a time we only need worry about that day. So "Event" does not represent the whole event but just the part of the event on that day?

danrabbit:
@jeremypw my long term goal is removing the Event class completely. It's only used in order to build EventRow so it adds a lasagna layer for no real reason. #166 and #167 together get us to a place where we can remove it. This is just kind of a transitional step. If #166 is merged first, I should be able to update this branch to completely remove the Event class

src/Widgets/EventRow.vala
```diff
@@ -18,15 +18,23 @@
  */
 
 public class DateTime.EventRow : Gtk.ListBoxRow {
-    public DateTime.Event cal_event { get; construct; }
+    public GLib.DateTime date { get; construct; }
+    public unowned ICal.Component component { get; construct; }
+
+    public GLib.DateTime start_time { get; private set; }
+    public GLib.DateTime end_time { get; private set; }
```
jeremypw: Since this can be null, should be `GLib.DateTime?`

PR #172, opened 2019-07-24
Fix calendar scroll

### LINKED ISSUE #38: Calendar scrolling lag
Months scrolling lags heavily after last update.
Moreover seems like the whole wingpanel freezes. I'm using `System Load Indicator` and it's stops for a while on calendar scroll.

I've got a PC that powerful enough:
- Quad-Core Intel® Core™ i5-3570K CPU @ 3.40GHz
- 12.0 GB memory

And it wasn't like this before update.
So it seems like a bug.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46846207-calendar-scrolling-lag?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57456330&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #38 

Import smooth-scrolling logic from wingpanel-indicator-sound to replace Util.on_scroll_event (). This gives a more even and controllable scrolling using mouse and touch.  Also respects natural scrolling settings.

### DIFF
--- a/src/Widgets/calendar/CalendarView.vala
+++ b/src/Widgets/calendar/CalendarView.vala
@@ -185,10 +185,18 @@ public class DateTime.Widgets.CalendarView : Gtk.Grid {
 
         double dir = 0.0;
         bool natural_scroll;
+        var source_device = e.get_source_device ();
         var event_source = e.get_source_device ().input_source;
-        if (event_source == Gdk.InputSource.MOUSE) {
+
+        /* Fallback to device name to try to detect a touch device that reports itself as a mouse */
+        bool is_touchpad = (event_source == Gdk.InputSource.TOUCHPAD ||
+                           source_device.get_name ().up ().contains ("TOUCH"));
+
+        bool is_mouse = !is_touchpad && (event_source == Gdk.InputSource.MOUSE);
+
+        if (is_mouse) {
             natural_scroll = natural_scroll_mouse;
-        } else if (event_source == Gdk.InputSource.TOUCHPAD) {
+        } else if (is_touchpad) {
             natural_scroll = natural_scroll_touchpad;
         } else {
             natural_scroll = false;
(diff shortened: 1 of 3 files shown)


jeremypw:
@danrabbit . I fixed the scroll direction.  However I find this much easier to control than master because the calendar *always* scrolls when you swipe and with the same delta per month.  If you swipe slowly then the calendar scrolls slowly, whereas if you swipe faster the calendar scrolls faster.  In master , if you swipe too slowly then the calendar never scrolls and the delta per month depends on the speed of swipe.

I can increase the delta per month (lines 282 , 284) from 2.0 to 3.0 or higher  if you think the controllability is insufficient.

jeremypw:
I have pushed a new version which puts the logic into CalendarView.vala and reduces scroll sensitivity to swiping.   Also, scroll events are ignored during transitioning.  Mouse scrolling events are adjusted so that one month per click of scroll wheel is maintained.

jeremypw:
I have had to add a "hack" to deal with touch devices that are incorrectly reported as Gdk.InputSource.MOUSE (causing too fast scrolling).

danrabbit:
Requesting a review from @tintou here since I believe he wrote the scroll handling for the applications menu

jeremypw:
OK, good idea.  The code here was originally intended for controlling a scale widget.  There may well be a better solution for switching a stack by scrolling.

tintou:
I'd prefer to wait until we get proper gesture support in the Gtk stack with GtkGesture (something like https://valadoc.org/gtk+-3.0/Gtk.GestureSwipe.html )

jeremypw:
OK, the current behaviour is usable 'til then.

PR #188, opened 2019-09-21
Change margin on day numbers so that they are centered even with dots

lainsce:
If a day has no events, it shouldn't reserve space for dots, as such, this PR solves this problem which also as a bonus helps the selection circle have days in the middle correctly.

Screenshot:
Screenshot from 2019-09-21 00 38 09


src/Widgets/calendar/GridDay.vala
```diff
@@ -115,6 +115,8 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
                         event_dots[event_uid] = event_dot;
 
                         event_grid.add (event_dot);
+                        grid.attach (event_grid, 0, 1);
```
tintou: You're calling attach several times here, you can check `if (event_grid.get_parent () != null)` before calling it

src/Widgets/calendar/GridDay.vala
```diff
@@ -115,6 +115,8 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+                        grid.margin_top = 6;
```
tintou: Put this at grid creation time, not here

src/Widgets/calendar/GridDay.vala
lainsce: Done. See the newest commit. :smile:

src/Widgets/calendar/GridDay.vala
lainsce: Done.

lainsce:
Review points addressed, any more stuff to do before merge?

danrabbit:
Week numbers are no longer aligned in this branch

Screenshot from 2019-09-23 10 14 48@2x

lainsce:
Oops, that'll be fixed as well as soon as I can. I always forget about that part of the UI. :laughing:

lainsce:
Week numbers aligned.

Screenshot from 2019-09-23 15 31 41

src/Widgets/calendar/GridDay.vala
```diff
@@ -56,21 +57,30 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
 
         label = new Gtk.Label (null);
 
+        event_dots = new Gee.HashMap<string, Gtk.Widget> ();
```
danrabbit: Why did you move these 4 lines here? Is it necessary?

src/Widgets/calendar/GridDay.vala
```diff
@@ -56,21 +57,30 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+
+        model.events_added.connect (add_event_dots);
+        model.events_removed.connect (remove_event_dots);
+
         event_grid = new Gtk.Grid ();
         event_grid.halign = Gtk.Align.CENTER;
         event_grid.height_request = 6;
 
-        var grid = new Gtk.Grid ();
```
danrabbit: It doesn't look like you're accessing this grid from outside of the `construct` block. Why did you make it available to the whole class?

src/Widgets/calendar/GridDay.vala
```diff
@@ -56,21 +57,30 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
+        grid = new Gtk.Grid ();
         grid.halign = grid.valign = Gtk.Align.CENTER;
         grid.attach (label, 0, 0);
-        grid.attach (event_grid, 0, 1);
+
+        if (event_grid.get_parent () == null) {
```
danrabbit: Do we have reason to believe that event_grid already has a parent at construction time? It doesn't look like there's any other place where event_grid is attached or added to another widget

lainsce: It seems so. If I set it to `event_grid.get_parent () != null` then no dots are shown for any events.

src/Widgets/calendar/GridDay.vala
lainsce: So I could use it on the add_event_dots method.

src/Widgets/calendar/GridDay.vala
lainsce: Needed to capture the events into the hashmap before doing the grid stuff for the days.

lainsce:
Hopefully my explanations make sense :thinking:

src/Widgets/calendar/GridDay.vala
danrabbit: Right, because when this class is constructed `event_grid` should not yet have a parent and there's nowhere else in this class where `event_grid` is attached or added to the grid. So this code might as well say `if (1 == 1)`

lainsce: I see. I'll fix it right away to not have the `if` block.

lainsce:
@danrabbit Anything more that is needed here?

lainsce:
@tintou Have made fixes for your review points as well before Dan's review point fixes.

lainsce:
> I'm not sure this is an improvement since it just makes days with events look off center instead of days with no events looking off center.

I'm not sure why that's happening over there, maybe HiDPI; see this screenshot where I use the PR, which has the focus circle and day text here in perfect alignment as it should be:
Screenshot from 2019-10-22 14 08 39

lainsce:
I also changed PR name to reflect current PR codebase

danrabbit:
Right so now the numbers are centered, but the whole area (the numbers and the dots) are closer to the bottom than the top.

So really this PR doesn't smartly do anything it all. It moves margin from one side to the other.

I don't think it would require anything hacky at all to draw the background around the number only. It looks like we'd need to remove some CSS we currently add to change the dots to white and then add something like:

```
.circular {
    background: transparent;
}

.circular:selected label {
    background: @selected_bg_color;
    border-radius: 50%;
    min-height: 1.5em;
    min-width: 1.5em;
    padding: 1px;
}
```

lainsce:
I'll use that css for the next commit

Dirli:
https://github.com/Dirli/wingpanel-indicator-datetime/commit/cb3c803eafcf7198bbcb23318dbeaae670e5fd27
Maybe this?

danrabbit:
Closing in favor of #222

Issue #206, opened 2020-03-28
No event dots since added libecal-2.0 support


worldofpeace:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
I add some events to my calendar on the same day and wingpanel-indicator-datetime doesn't show any dots counting those events.

### To Reproduce
1. Build wingpanel-indicator-datetime with libecal-2.0
2. Add events with elementary-calendar
3. Open wingpanel-indicator-datetime and navigate to that day
4. See no dots on date

### Expected behavior

I would see some pretty dots

### Screenshots or screen recordings
![Screenshot from 2020-03-28 18 57 57](https://user-images.githubusercontent.com/28888242/77835740-1c8e3880-7126-11ea-88f4-9603335be962.png)

### Logs
Running wingpanel with `G_MESSAGES_DEBUG=all` yielded no errors when selecting wingpanel-indicator-datetime.

## Platform Information

- OS: NixOS
- OS Version: 20.09 git
This is using a GNOME 3.34 stack. (not for long though, will see if I can still reproduce when that upgrade is merged).

- [ ] I'm using the latest version from git that I've manually compiled
- [x] I'm using the latest released stable version

J0Ar:
Event dots are also gone on elementaryOS. 🤷‍♂️️

worldofpeace:
@J0Ar Very strange. Could you query with dpkg (or apt, can't remember) what version of evolution-data-server is at?

J0Ar:
@worldofpeace I tried an older installation of elementaryOS, where the indicator was at version `2.2.1+r555+pkg22~ubuntu5.1.1` and there the event dots are still working.
 
But after the upgrade to `2.2.2+r580+pkg23~ubuntu5.1.2.1` and a logout > login the dots are gone.

The **evolution-data-server** was always at…
```
apt-cache show evolution-data-server
Package: evolution-data-server
Architecture: amd64
Version: 3.28.5-0ubuntu0.18.04.2
```

worldofpeace:
@J0Ar The only change, which isn't in the release notes, is https://github.com/elementary/wingpanel-indicator-datetime/commit/3ccd05d611e6dd5274a03f061ba1b5e13d6fe0cf.
That allows it to build with latest libecal. What likely happened here (a guess), is the code path that is for the event dots is broken in a libecal-2.0 version, and maybe the conditional compilation isn't right and that code is used for 3.28 as well.

[...]

PR #207, opened 2020-04-16
evolution optional dependency

Dirli:
I really did not like that the evolution-data-server was a mandatory dependency, I tried to make it an optional dependency, for those who do not need it) I think it worked. Several issues were corrected along the way.
issue #206 fix too


danrabbit:
Hey thanks for your branch!

I'm not certain if it makes sense to carry these changes since other parts of Pantheon use EDS as well. What is the value in adding a conditional compile for Evolution here?

As it stands, there's no way to review or merge this branch. It has too many unrelated changes. A diff of +1,354 −1,205 is not realistic to review. If you'd like to propose several changes (like reorganizing files, etc), please use multiple branches

Dirli:
You very quickly learned all the changes) unfortunately in smaller chunks long and more labor-intensive

PR #209, opened 2020-04-24
EventRow: Fix appearance in dark style

### LINKED ISSUE #208: Events illegible in dark style
Not sure how this should look:

Screenshot from 2020-04-23 17 43 05@2x

danrabbit:
Fixes #208 

## BEFORE
Screenshot from 2020-04-23 17 43 05@2x
Screenshot from 2020-04-23 17 53 51@2x

## AFTER
Screenshot from 2020-04-23 17 51 09@2x
Screenshot from 2020-04-23 17 51 23@2x

### DIFF
--- a/data/EventRow.css
+++ b/data/EventRow.css
@@ -1,5 +1,5 @@
 .event {
-    background: alpha(@colorAccent, 0.15);
+    background: mix(@colorAccent, shade (#fff, 0.96), 0.8);
     border-radius: 3px;
     padding: 6px;
 }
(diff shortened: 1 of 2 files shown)


cassidyjames:
@danrabbit this make it legible, but it's super bright. I think ideally we would have them look a bit dimmer/darker in the dark style (kinda like the badges on releases.elementary.io). But also, this works as a starting point if we wanna start here.

danrabbit:
Yeah maybe we can figure out a more complex solution later and take this as a hot fix so it’s not totally broken?

lainsce:
What if the dark mode styling was the text and icons color for the bg, and vice versa? More contrasty that way and less bright.

PR #216, opened 2020-05-18
Minor refactoring and improve the performance and reliability of event dots.

ikarosdev:
- Take some of the refactoring work done by @Dirli.
- Increase performance of indicator while scrolling. With no dots in the indicator, the scrolling is smooth between months. If dots are present, the performance takes a hit and is sluggish while moving. This seems to fix that issue.
- Improve reliability of dots showing up. In the case of 3 dots, sometimes the indicator would require a couple openings to show. In the case of 3 dots, oftentimes the indicator would not show 3 dots upon startup. This seems to make event dots appear more reliably for the user.

### DIFF
--- a/src/Widgets/calendar/Grid.vala
+++ b/src/Widgets/calendar/Grid.vala
@@ -26,6 +26,7 @@ namespace DateTime.Widgets {
  */
     public class Grid : Gtk.Grid {
         public Util.DateRange grid_range { get; private set; }
+        private static Widgets.CalendarModel model;
 
         /*
          * Event emitted when the day is double clicked or the ENTER key is pressed.
@@ -40,6 +41,7 @@ namespace DateTime.Widgets {
         private Gtk.Revealer[] week_labels;
 
         construct {
+            model = Widgets.CalendarModel.get_default ();
             header_labels = new Gtk.Label[7];
             for (int c = 0; c < 7; c++) {
                 header_labels[c] = new Gtk.Label (null);
@@ -64,6 +66,8 @@ namespace DateTime.Widgets {
             data = new Gee.HashMap<uint, GridDay> ();
             events |= Gdk.EventMask.SCROLL_MASK;
             events |= Gdk.EventMask.SMOOTH_SCROLL_MASK;
+            model.events_added.connect (add_event_dots);
+            model.events_removed.connect (remove_event_dots);
         }
 
         private void on_day_focus_in (GridDay day) {
@@ -259,6 +263,36 @@ namespace DateTime.Widgets {
 
         private uint day_hash (GLib.DateTime date) {
             return date.get_year () * 10000 + date.get_month () * 100 + date.get_day_of_month ();
+        }
+        
+                
+        public void add_event_dots (E.Source source, Gee.Collection<ECal.Component> events) {
+            foreach (var component in events) {
+            
+                unowned ICal.Component ical = component.get_icalcomponent ();
+                ICal.Time? start_time = ical.get_dtstart ();
+                time_t start_unix = start_time.as_timet ();
+                var t = new GLib.DateTime.from_unix_utc (start_unix);
+                var d_hash = day_hash (t);
+                if (data.has_key (d_hash)) {
+                    data[d_hash].add_dots (source, component.get_icalcomponent ());
+                }
+            }
+        }
+
+        public void remove_event_dots (E.Source source, Gee.Collection<ECal.Component> events) {
+            foreach (var component in events) {
+                unowned ICal.Component ical = component.get_icalcomponent ();
+                var event_uid = ical.get_uid ();
+
+                ICal.Time? start_time = ical.get_dtstart ();
+                time_t start_unix = start_time.as_timet ();
+                var t = new GLib.DateTime.from_unix_utc (start_unix);
+                var d_hash = day_hash (t);
+                if (data.has_key (d_hash)) {
+                    data[d_hash].remove_dots (event_uid);
+                }
+            }
         }
 
     }
--- a/src/Widgets/calendar/Grid.vala
+++ b/src/Widgets/calendar/Grid.vala
@@ -40,8 +40,11 @@ namespace DateTime.Widgets {
         private Gtk.Label[] header_labels;
         private Gtk.Revealer[] week_labels;
 
-        construct {
+        static construct {
             model = Widgets.CalendarModel.get_default ();
+        }
+
+        construct {
             header_labels = new Gtk.Label[7];
             for (int c = 0; c < 7; c++) {
                 header_labels[c] = new Gtk.Label (null);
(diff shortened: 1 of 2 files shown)


src/Widgets/calendar/Grid.vala
```diff
@@ -40,6 +41,7 @@ namespace DateTime.Widgets {
         private Gtk.Revealer[] week_labels;
 
         construct {
+            model = Widgets.CalendarModel.get_default ();
```
danrabbit, line 44: If we want the model to be static, probably we should create it in a `static construct {}` block

src/Widgets/calendar/Grid.vala
```diff
@@ -260,6 +264,36 @@ namespace DateTime.Widgets {
         private uint day_hash (GLib.DateTime date) {
             return date.get_year () * 10000 + date.get_month () * 100 + date.get_day_of_month ();
         }
+        
+                
+        public void add_event_dots (E.Source source, Gee.Collection<ECal.Component> events) {
+            foreach (var component in events) {
+
```
danrabbit, line 271: Extra whitespace here :)

src/Widgets/calendar/GridDay.vala
```diff
@@ -43,8 +42,7 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
     }
 
     static construct {
-        model = Widgets.CalendarModel.get_default ();
-
+
```
danrabbit, line 45: Extra whitespace here too

src/Widgets/calendar/Grid.vala
```diff
@@ -260,6 +264,36 @@ namespace DateTime.Widgets {
...
             return date.get_year () * 10000 + date.get_month () * 100 + date.get_day_of_month ();
         }
+        
+                
+        public void add_event_dots (E.Source source, Gee.Collection<ECal.Component> events) {
+            foreach (var component in events) {
+            
+                unowned ICal.Component ical = component.get_icalcomponent ();
+                ICal.Time? start_time = ical.get_dtstart ();
+                time_t start_unix = start_time.as_timet ();
```
danrabbit, line 274: Since `start_time` is nullable, this probably needs a null check

src/Widgets/calendar/Grid.vala
```diff
@@ -260,6 +264,36 @@ namespace DateTime.Widgets {
...
+                var t = new GLib.DateTime.from_unix_utc (start_unix);
+                var d_hash = day_hash (t);
+                if (data.has_key (d_hash)) {
+                    data[d_hash].add_dots (source, component.get_icalcomponent ());
```
danrabbit, line 278: You already have `component.get_icalcomponent ()` cached as `ical` right? Probably don't need to get it again

src/Widgets/calendar/Grid.vala
```diff
@@ -260,6 +264,36 @@ namespace DateTime.Widgets {
         private uint day_hash (GLib.DateTime date) {
             return date.get_year () * 10000 + date.get_month () * 100 + date.get_day_of_month ();
         }
+        
+                
+        public void add_event_dots (E.Source source, Gee.Collection<ECal.Component> events) {
```
danrabbit, line 269: This seems to only be used internally right? So it can probably be `private`

src/Widgets/calendar/Grid.vala
```diff
@@ -260,6 +264,36 @@ namespace DateTime.Widgets {
...
+                time_t start_unix = start_time.as_timet ();
+                var t = new GLib.DateTime.from_unix_utc (start_unix);
+                var d_hash = day_hash (t);
+                if (data.has_key (d_hash)) {
+                    data[d_hash].add_dots (source, component.get_icalcomponent ());
+                }
+            }
+        }
+
+        public void remove_event_dots (E.Source source, Gee.Collection<ECal.Component> events) {
```
danrabbit, line 283: Same as above, this can probably be `private`

src/Widgets/calendar/Grid.vala
```diff
@@ -260,6 +264,36 @@ namespace DateTime.Widgets {
...
+            foreach (var component in events) {
+                unowned ICal.Component ical = component.get_icalcomponent ();
+                var event_uid = ical.get_uid ();
+
+                ICal.Time? start_time = ical.get_dtstart ();
+                time_t start_unix = start_time.as_timet ();
```
danrabbit, line 289: Same as above, probably needs a null check

src/Widgets/calendar/GridDay.vala
```diff
@@ -84,53 +82,49 @@ public class DateTime.Widgets.GridDay : Gtk.EventBox {
...
-            unowned ICal.Component ical = component.get_icalcomponent ();
-            var event_uid = ical.get_uid ();
-            var dot = event_dots[event_uid];
-            if (dot != null) {
-                dot.destroy ();
-                event_dots.remove (event_uid);
-            }
+        var dot = event_grid.get_children ();
+        if (dot.length () > 0) {
+            dot.nth_data (0).destroy ();
```
danrabbit, line 127: This seems less efficient and less accurate than using the hashmap as we have it in master. There's no guarantee that the first dot is the correct one to remove, which means you could have dots with colors that don't match what's actually on that day

ikarosdev:
Thank you for the feedback. I'll clean it up later today when I have some time. I'm not a programmer (just a hobbyist), so I'm learning what's "proper" and not. As for the null checks, does that mean you want a conditional there (i.e. if (!null))?

danrabbit:
Definitely a good job cherrypicking then for not considering yourself a programmer!

Right, so if we end up fetching a `null` start time then we won't be able to convert it, hash it, etc because it doesn't exist. So we'll likely want something there that says if `ical.get_dtstart ()` returns `null` then we just stop there and skip to the next component

Dirli:
https://valadoc.org/libical-glib/ICal.Component.get_dtstart.html
is an null value returned?

src/Widgets/calendar/GridDay.vala
Dirli: Currently, all your points are depersonalized. Just point with a single class, why should they be personalized? If you try to add, say, 6 events, only 3 will be added. Now that you delete any of the added 3 events, what happens? correctly, the point will be deleted. As a result, you will get 2 points and 5 events.
The code that you reject at least checks whether it is time to delete a point, or only reduce the number of events for this point

ikarosdev:
Alright, so be gentle (seriously, I have no idea what I'm doing). But I added the conditionals you requested. I believe I returned GridDay back to how the function works upstream and I cleaned up any rift-raft whitespace complaints.

ikarosdev:
I just realized I borked this. I’ll have to commit a fix.

src/Widgets/calendar/GridDay.vala
ikarosdev: What's a better solution for this? Should I run an iterator, search the list and destroy the element that way?

Dirli: I outlined my opinion above. I hope someone from the developers will voice what logic they laid. I assume that I am wrong, because I do not use this functionality)

ikarosdev: @Dirli and @danrabbit. I think I understand what @danrabbit is saying. I still feel an arraylist is faster than the hashmap implementation. The problem is the dots are color coded for different types of events (I've never noticed this before). So, if we always remove the first dot, that might not be the right one to remove.

ikarosdev: @Dirli I'm not a developer so I'm struggling here. Is there a way to search the arraylist and remove an element from that list based on what event_uid is?

Dirli: Yes, the colors were a surprise to me (while fix issue, I did not come across them). My solution needs to be rewritten to work correctly. Leaving the discussion)

ikarosdev:
I'm going to close this PR. While on the surface this seems to improve reliability and function. This kills certain functionality of the indicator.

danrabbit:
Thank you for starting this and the idea! I submitted a branch that should have the same result here: https://github.com/elementary/wingpanel-indicator-datetime/pull/217

I didn't end up including the null checks since as Dirli pointed it, the API doc doesn't seem to indicate that getting start time should be nullable

ikarosdev:
@danrabbit Thanks, man. I'm hoping to become a more frequent contributor around these parts.

danrabbit:
@ikarosdev that would be great! I appreciate your effort here. Thank you for submitting this PR and getting the conversation going

ikarosdev:
@danrabbit Sorry if this isn't the place, but just to clarify (mostly because I've been working on this all day), but I assumed using ArrayList vs HashMap was working "faster," which improved overall performance of the indicator. When I went with the changes @Dirli came up with, I found myself in a world of hurt trying to figure out the right calls (I think I'm using this word correctly) work for ArrayList vs HashMap and I was having to dig through Vala docs. However, the "fix" here is less CalendarModels? In simplest terms, can you explain that?

danrabbit:
@ikarosdev right so what’s happening here is we have one grid per month (which honestly this class should be renamed so that’s clear) and inside that grid we have one GridDay per day.

In master, each day is connecting to the add and remove events from the calendar model. So per month that’s typically 30 times we’re running this method that has to figure out if the event is meant for this month and then add it or not.

What this refactor does is instead puts those connections in the month grid and the month grid will figure out which days to add the events to. So we’re doing far less work each time we build a new month grid.

PR #225, opened 2020-06-15
CalendarStore: Support Events and Tasks

marbetschar:
This pull request introduces an all new `CalendarStore` object which intents to merge the features of `CalendarModel` and `TaskModel` with the goal to provide a unified API and eventually replace both.

For events use `CalendarStore.get_event_store ()`, for tasks use `CalendarStore.get_task_store ()`.

This PR fixes https://github.com/elementary/wingpanel-indicator-datetime/issues/211.


marbetschar:
Added support to display tasks together with events in the datetime indicator using the new CalendarStore object:

Screenshot from 2020-06-16 10 05 18

marbetschar:
We might want to use another symbolic icon for tasks than `alarm-symbolic` - something that makes the relation to tasks more obvious. Probably just a checkmark in a box.

As far as I remember there exists one, but I don't find it anymore. @cassidyjames can you help?

marbetschar:
@danrabbit just seen there is a project ([Wingpanel 3](https://github.com/elementary/wingpanel/projects/1)) which is targeted towards odin. Does it make sense to add this PR to this project (cause Tasks is also scheduled for odin)?

marbetschar:
**UPDATE:** [According to the discussion in Calendar](https://github.com/elementary/calendar/pull/588#issuecomment-692160733), this PR wont be merged because we won't create an abstract super class and therefore keep code in Tasks and Calendar separated.

PR #228, opened 2020-07-08
Prevent loading DateTime outside session

worldofpeace:
Noticed this in https://github.com/elementary/greeter/pull/450 https://github.com/elementary/greeter/pull/450#issuecomment-655130041. I tested that this indicator doesn't show in the greeter now in a vm.

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -225,8 +225,14 @@ public class DateTime.Indicator : Wingpanel.Indicator {
     }
 }
 
-public Wingpanel.Indicator get_indicator (Module module) {
+public Wingpanel.Indicator get_indicator (Module module, Wingpanel.IndicatorManager.ServerType server_type) {
     debug ("Activating DateTime Indicator");
+
+    if (server_type != Wingpanel.IndicatorManager.ServerType.SESSION) {
+        debug ("Wingpanel is not in session, not loading DateTime");
+        return null;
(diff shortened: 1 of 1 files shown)


worldofpeace:
cc @danrabbit

src/Indicator.vala
```diff
@@ -225,8 +225,14 @@ public class DateTime.Indicator : Wingpanel.Indicator {
     }
 }
 
-public Wingpanel.Indicator get_indicator (Module module) {
+public Wingpanel.Indicator get_indicator (Module module, Wingpanel.IndicatorManager.ServerType server_type) {
```
Dirli, line 228: Wingpanel.Indicator?

worldofpeace: Hi, could you be a little more specific with what is needed here?

Dirli: if you return not only a value, but also a null, you must specify this (return_value?)

Dirli: я получаю такое предупреждение без "?"
`warning: `null' incompatible with return type `Wingpanel.Indicator'`

worldofpeace, line 228:
```suggestion
public Wingpanel.Indicator? get_indicator (Module module, Wingpanel.IndicatorManager.ServerType server_type) {
```

Issue #210, opened 2020-04-26
Use Hdy.Paginator


danrabbit:
For smooth 1:1 swiping on the calendar

Dirli:
I don't want to trash, so I'll write here.
PR (232) not very good. Yes, it implements the required functionality, but it brings problems with it.
![screen1](https://user-images.githubusercontent.com/15650716/95398666-14058e80-090f-11eb-9d1b-640500513a99.png)
this I switched back to 2000. CalendarGrid accumulate for some reason, and if you switch for a very long time, the memory will suffer. but even worse, when you close the widget, the array of CalendarGrid is not cleared, so it all remains in memory.
![screen2](https://user-images.githubusercontent.com/15650716/95399014-e0773400-090f-11eb-90e3-8a5519a33329.png)
also, each month switch generates a lot of unnecessary actions, at each iteration, there are 4 changes to the model, which entails almost complete redrawing of everything related to events.
it seems to me that it would be more logical to leave 3 months (previous, current, next). delete dynamically everything else. the problem of memory consumption will be solved.
PS stop using the global scope and singleton, with this approach, the future is very doubtful. supporting such code is difficult. this project has much less code (compared to Music), but it is also very difficult to read.

danrabbit:
@Dirli can you open a new issue report about this please?

PR #232, opened 2020-09-22
Use Hdy.Carousel

fs3o5:
Adds  #210

### DIFF
--- a/src/Widgets/calendar/CalendarView.vala
+++ b/src/Widgets/calendar/CalendarView.vala
@@ -104,7 +104,7 @@ public class DateTime.Widgets.CalendarView : Gtk.Grid {
         attach (carousel, 0, 1, 2);
 
         left_button.clicked.connect (() => {
-            carousel.switch_child ((int) carousel.get_position () + -1, carousel.get_animation_duration ());
+            carousel.switch_child ((int) carousel.get_position () - 1, carousel.get_animation_duration ());
         });
 
         right_button.clicked.connect (() => {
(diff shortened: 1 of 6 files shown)


danrabbit:
Don't forget to update CI :) You'll need to add the libhandy package to the build step

src/Widgets/calendar/CalendarView.vala
```diff
@@ -121,16 +178,44 @@ public class DateTime.Widgets.CalendarView : Gtk.Grid {
...
-        if (!start.equal (calmodel.month_start)) {
-            calmodel.month_start = start;
-        }
-        sync_with_model ();
+        if (start.equal (start_month)) {
+            position -= rel_postion;
+            carousel.switch_child (position, carousel.get_animation_duration ());
+        } else {
+            /*reset Carousel if center_child != the grid of the month of today*/
+            carousel.set_no_show_all (true);
```
danrabbit, line 190: elementary code style is to prefer properties :)
```suggestion
            carousel.no_show_all = true;
```

src/Widgets/calendar/CalendarView.vala
```diff
@@ -121,16 +178,44 @@ public class DateTime.Widgets.CalendarView : Gtk.Grid {
...
+            foreach (Gtk.Widget grid in carousel.get_children ()) {
```
danrabbit, line 191: I think we can use an unowned reference here
```suggestion
            foreach (unowned Gtk.Widget grid in carousel.get_children ()) {
```

danrabbit (changes requested):
Thank you for working on this! I know this code is kind of a mess. But it looks like your branch is working very smoothly!

I've left a few inline comments about code style, but as far as I can tell this works as expected. Great job!

danrabbit (approved):
Great job! This works super smoothly

PR #236, opened 2020-10-09
EventRow: Rely on stylesheet for colors

danrabbit:
Combined with https://github.com/elementary/stylesheet/pull/852 , relies on the theme for background and foreground colors. Fixes issues with dark style


cassidyjames (approved):
Looks :fire:

PR #240, opened 2020-11-13
Use First Day of Week Setting from Switchboard Locale Plug

lainsce:
Using https://github.com/elementary/switchboard-plug-locale/pull/127, change the order of weekdays.

Currently non-live, needs to killall wingpanel and bring it back for this setting to take effect.


lainsce:
Would love some help on making it live as change occurs.

lainsce:
Technically sound code, but I don't know why it doesn't change live.

lainsce:
It actually updates, but only after 10 months going forwards, or 2 months going back. Month grid generation is thus flawed.

lainsce:
Ready to review.

src/Widgets/calendar/CalendarView.vala
```diff
@@ -79,6 +79,31 @@ public class DateTime.Widgets.CalendarView : Gtk.Grid {
         right_grid.update_weeks (calmodel.data_range.first_dt, calmodel.num_weeks);
         calmodel.change_month (-1);
 
+        CalendarModel.get_default ().locale_settings.changed.connect (() => {
+            // If First Day setting changes, reset every grid.
+            carousel.no_show_all = true;
```
jeremypw, line 84: Better to construct `carousel` before using it :stuck_out_tongue:

lainsce: There are more problems with this PR, but I think those are out of scope and could be handled in a newer PR.

But thanks for finding this.

lainsce:
:)

jeremypw:
Converted to draft as it cannot be merged or properly tested before the switchboard PR is merged.

PR #248, opened 2021-02-07
Add Weather info

danrabbit:
Todo:
- [ ] add an icon to the panel?
- [ ] Change background color with conditions
- [ ] Figure out what to do in failure cases
- [ ] Decide which information to show
- [ ] Figure out what to do when it's not today

Screenshot from 2021-02-07 12 52 17@2x

### DIFF
--- a/data/EventRow.css
+++ b/data/EventRow.css
@@ -7,6 +7,7 @@
 .event label,
 .event image {
     color: @selected_fg_color;
+    -gtk-icon-palette: error @selected_fg_color, success @selected_fg_color, warning @selected_fg_color;
 }
 
 .event .title {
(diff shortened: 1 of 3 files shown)


tintou:
Evolution Data Server already has a "Weather" calendar source that you can use natively without manually linking on libgweather

4jNsY6fCVqZv:
How does the privacy of users stay protected in the context of this feature? Will it be enabled by default?

hanaral:
> How does the privacy of users stay protected in the context of this feature? Will it be enabled by default?
I presume this needs UX review anyway, like requesting access location before doing anything

amka:
Even windows 11 shows the weather in the tray. That said the point "add an icon to the panel" required 😉

andirsun:
Example of icon in panel. 
Screenshot from 2021-08-17 11-15-00

PR #251, opened 2021-03-01
Fixed an issue where focus style was not reset to today

marbetschar:
Whenever we select another day than "today" and then click on the calendar icon to reset to "today", the rounded styling of the previously selected day is not removed - despite the fact, that the shown events are from today.

This PR fixes this issue.


davidmhewitt (approved):
Makes sense with no obvious regressions found

PR #254, opened 2021-03-30
Fix #253: Reset indicator time on NewUser, NewSeat and NewSession events

### LINKED ISSUE #253: Highlighted 'today' day number falls out of sync over time
## What Happened

Today is the 14th of March, Sunday - however when I open the indicator it auto-selects yesterday(the 13th of March, Saturday) and when I manually focus another day the 13th is still in bold and accent coloured. This is pretty misleading, and presumably  it's because I suspend my computer overnight. 

## Expected Behavior

The date is shown correctly every time

## Steps to Reproduce

1. Suspend your device between date changes 
2. Open indicator and note which day is marked as today

## Platform Information

elementary OS 6 Daily

hotsezus:


davidmhewitt:
I don't think this fixes #253 as the `TimeManager` class is not used for the calendar widget.

hotsezus:
> I don't think this fixes #253 as the `TimeManager` class is not used for the calendar widget.

Got it. Will look further to fix a bug

Dirli:
DateTime.Widgets.Grid:
`public void set_focus_to_today () {
     if (grid_range == null) {
         return;
     }
     Gee.List<GLib.DateTime> dates = grid_range.to_list ();
     for (int i = 0; i < dates.size; i++) {
         var date = dates[i];
         GridDay? day = data[day_hash (date)];
         if (day != null && day.name == "today") {
             day.grab_focus_force ();
             return;
         }
     }
 }`
Think about it, it looks like that's where the problem is. I haven't checked yet, but it seems that you need to abandon the reference to "name=today", and directly access the desired day (via day_hash), and so that the old day is not lost, write it to a variable.
Perhaps tomorrow in my fork I will fix this error, you can do it by analogy.

Dirli:
The problem turned out to be a little wider than I originally thought. There are two ways to solve it, in my opinion:
1. the simple way is to reset the carousel when opening the widget. the carousel will be built again and correctly determine the current day
2.  a difficult way (in my opinion, more correct), to determine the current day at the level of the CalendarView (now you have at the level of CalendarGrid's). This will take much more time, unfortunately, I am not ready to transfer my conjectures to the main branch, but you can take this commit as a vector: https://github.com/Dirli/wingpanel-indicator-datetime/commit/93f5a144dde28e2f9639e81bc61a4a3d03be00ff

PR #261, opened 2021-05-08
Define log domain

peteruithoven:
Fixes: #260


jeremypw:
I'll wait a bit to make sure there is no pushback from my previous related approval/merge.

PR #287, opened 2021-11-06
Free libical cache before exiting the application

tintou:
This allows to not technically leak memory.


src/Widgets/calendar/Util.vala
```diff
@@ -71,22 +73,25 @@ namespace Util {
                 return new GLib.TimeZone (tzid);
             }
         }
-
-        //FIXME See https://github.com/libical/libical/pull/513
         // If tzid fails, try ICal.Time.get_timezone ()
-        ICal.Timezone* timezone = date.get_timezone ();
+        unowned ICal.Timezone? timezone = null;
+        if (timezone == null && date.get_timezone () != null) {
```
jeremypw, line 78: It seems unnecessary to test for `null` here as you have just set `timezone` to `null` - or am I missing something?

tintou: You're right, applied the same solution as in https://github.com/elementary/calendar/pull/722

danrabbit (approved):
Approving since this seems to be the same as in https://github.com/elementary/calendar/pull/722

PR #291, opened 2021-12-02
ComponentRow: Add "due by" for tasks

danrabbit:
Since this is a time you need to take action by, and not a time when something happens, I think it makes it more clear to add a little bit of a distinction beyond the icon here


marbetschar:
LGTM - however, because I merged the main branch into this, I'm no longer able to approve and merge :smile:

PR #292, opened 2021-12-02
Indicator: launch Tasks when clicking rows

danrabbit:
I'm not sure Tasks has a way to launch to the right task list?


marbetschar:
@danrabbit it has not (yet).

marbetschar (approved):

Issue #270, opened 2021-08-17
Add Button for adding event


snowparrot:
## Problem

I just want to add fast an event on the wing panel.

## Proposal

Add a button on the indicator so I can fast an event.

almir-lima:
One button [ + ] to call the App Calendar, it would be nice.

CodingSpiderFox:
Regarding the design, I have an idea: 
The date & time overlay that opens when clicking the current time in the middle of wingpanel could be extended by having a third column where quick action buttons like "Add event", "Add alarm", "Start Timer" could be added

PR #299, opened 2022-02-01
#270: Add button for adding event

benjaminkitt:
Addresses issue #270, Add button for adding event.

calendar_new_event

This PR is blocked by https://github.com/elementary/calendar/pull/742, which addresses source calendars not loading prior to the display of the new event dialog when loading calendar using the `--add-event` flag.

While a Dbus solution to this request is preferable (as noted in the pre-existing TODO), I felt that was beyond the scope of this PR as a Bitesize issue. I will explore a Dbus solution in a future PR. As it stands, this PR, along with the aforementioned PR on Calendar, does at least provide a good first step in providing the functionality.


src/Widgets/calendar/CalendarView.vala
```diff
@@ -241,8 +249,11 @@ public class DateTime.Widgets.CalendarView : Gtk.Grid {
     }
 
     // TODO: As far as maya supports it use the Dbus Activation feature to run the calendar-app.
-    public void show_date_in_maya (GLib.DateTime date) {
-        var command = "io.elementary.calendar --show-day %s".printf (date.format ("%F"));
+    private void open_maya_with_options (GLib.DateTime? day_to_show, bool add_event = false) {
```
benjaminkitt, line 252: Expanded the previous `show_date_in_maya` method to also allow for displaying the new event dialog. Public API remains the same and available.

PR #319, opened 2025-04-18
Port to GTK4

leolost2605:
Includes #320.

Merge that first and drop its commits since there were some more changes. Also ig we should squash this


src/Indicator.vala
```diff
@@ -84,14 +83,13 @@ public class DateTime.Indicator : Wingpanel.Indicator {
...
-            var scrolled_window = new Gtk.ScrolledWindow (null, null) {
+            var scrolled_window = new Gtk.ScrolledWindow () {
                 hscrollbar_policy = Gtk.PolicyType.NEVER,
                 child = component_listbox
             };
 
-            var settings_button = new Gtk.ModelButton () {
-                text = _("Date & Time Settings…")
-            };
+            var settings_button = new Gtk.Button.with_label (_("Date & Time Settings…"));
```
danirabbit, line 91: Should be PopoverMenuItem to make sure we set accessible role etc https://github.com/elementary/wingpanel/pull/607

src/Widgets/calendar/CalendarView.vala
```diff
@@ -201,12 +198,12 @@ public class DateTime.Widgets.CalendarView : Gtk.Grid {
         selected_date = today;
         if (start.equal (start_month) && !refresh) {
             position -= rel_postion;
-            carousel.switch_child (position, carousel.get_animation_duration ());
+            //todo as prep
```
danirabbit, line 201: Todo line here?

src/Widgets/calendar/CalendarView.vala
```diff
@@ -251,14 +246,14 @@ public class DateTime.Widgets.CalendarView : Gtk.Grid {
             var appinfo = AppInfo.create_from_commandline (command, null, AppInfoCreateFlags.NONE);
             appinfo.launch_uris (null, null);
         } catch (GLib.Error e) {
+            //todo: gtk4 prep
```
danirabbit, line 249: Another todo line. Need to connect to response here as well

danirabbit (changes requested):

leonardo-lemos (approved):

danirabbit (approved):
Let's go :rocket:

